PR #1, opened 2021-07-05
Add GitHub action for make and validate

jondea:
I've added a relatively simple make and validation workflow. Please feel free to leave any improvements or corrections.

On [the latest run](https://github.com/jondea/oomph-lib/runs/2985474578?check_suite_focus=true) it appears that clamped shell is failing, does anyone else know about this?
```
Validation directories where failures were detected:
 
   /home/runner/work/oomph-lib/oomph-lib/demo_drivers/shell/clamped_shell/Validation : [FAILURE NUMBER 1 ]
 
TOTAL FAIL COUNT: 1
 ```

### DIFF
--- a/.github/workflows/build-and-validate.yml
+++ b/.github/workflows/build-and-validate.yml
@@ -1,0 +1,17 @@
+name: Build and validate
+
+on: push
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install requirements
+      run: sudo apt-get install make automake libtool autoconf doxygen gfortran g++ 
+    - name: Build
+      run: ./non_interactive_autogen.sh
+    - name: Validate
+      run: ./validate.sh
--- a/.github/workflows/build-and-validate.yml
+++ b/.github/workflows/build-and-validate.yml
@@ -12,6 +12,6 @@ jobs:
     - name: Install requirements
       run: sudo apt-get install make automake libtool libtool-bin autoconf doxygen gfortran g++ 
     - name: Build
-      run: ./non_interactive_autogen.sh
+      run: ./non_interactive_autogen.sh -s
     - name: Validate
-      run: make check -k
+      run: make check -k -s
--- a/.github/workflows/build-and-validate.yml
+++ b/.github/workflows/build-and-validate.yml
@@ -14,4 +14,4 @@ jobs:
     - name: Build
       run: ./non_interactive_autogen.sh
     - name: Validate
-      run: ./validate.sh
+      run: make check -k
--- a/.github/workflows/build-and-validate.yml
+++ b/.github/workflows/build-and-validate.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Install requirements
-      run: sudo apt-get install make automake libtool autoconf doxygen gfortran g++ 
+      run: sudo apt-get install make automake libtool libtool-bin autoconf doxygen gfortran g++ 
     - name: Build
       run: ./non_interactive_autogen.sh
     - name: Validate
--- a/.github/workflows/build-and-validate.yml
+++ b/.github/workflows/build-and-validate.yml
@@ -9,15 +9,15 @@ jobs:
       matrix:
         include:
           - os: ubuntu-latest
-            install-script: sudo apt-get install
+            install-script: sudo apt-get install make automake libtool libtool-bin autoconf doxygen gfortran g++ 
           - os: macos-latest
-            install-script: brew install
+            install-script: brew install make automake libtool autoconf doxygen gfortran
 
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v2
     - name: Install requirements
-      run: ${{ matrix.install-script }} make automake libtool libtool-bin autoconf doxygen gfortran g++ 
+      run: ${{ matrix.install-script }}
     - name: Build
       run: ./non_interactive_autogen.sh -s
     - name: Validate
--- a/.github/workflows/build-and-validate.yml
+++ b/.github/workflows/build-and-validate.yml
@@ -22,7 +22,7 @@ jobs:
     - name: Install requirements
       run: ${{ matrix.install-script }}
     - name: Extra config
-      run: printf '${{ extra-config }}' >> config/configure_options/default
+      run: printf '${{ matrix.extra-config }}' >> config/configure_options/default
     - name: Build
       run: ./non_interactive_autogen.sh -s
     - name: Validate
--- a/.github/workflows/build-and-validate.yml
+++ b/.github/workflows/build-and-validate.yml
@@ -14,7 +14,7 @@ jobs:
             install-script: |
               brew install make automake libtool autoconf doxygen
               brew reinstall gfortran
-            extra-config: --enable-suppress-CGAL-build\n
+            extra-config: --enable-suppress-CGAL-build
 
     runs-on: ${{ matrix.os }}
     steps:
@@ -22,7 +22,7 @@ jobs:
     - name: Install requirements
       run: ${{ matrix.install-script }}
     - name: Extra config
-      run: printf '${{ matrix.extra-config }}' >> config/configure_options/default
+      run: printf '%s\n' '${{ matrix.extra-config }}' >> config/configure_options/default
     - name: Build
       run: ./non_interactive_autogen.sh -s
     - name: Validate
--- a/.github/workflows/build-and-validate.yml
+++ b/.github/workflows/build-and-validate.yml
@@ -15,3 +15,8 @@ jobs:
       run: ./non_interactive_autogen.sh -s
     - name: Validate
       run: make check -k -s
+    - name: Upload validation log file
+      uses: actions/upload-artifact@v2
+      with:
+        name: Validation log file
+        path: ./validation.log
--- a/.github/workflows/build-and-validate.yml
+++ b/.github/workflows/build-and-validate.yml
@@ -11,7 +11,9 @@ jobs:
           - os: ubuntu-latest
             install-script: sudo apt-get install make automake libtool libtool-bin autoconf doxygen gfortran g++ 
           - os: macos-latest
-            install-script: brew install make automake libtool autoconf doxygen gfortran
+            install-script: |
+              brew install make automake libtool autoconf doxygen
+              brew reinstall gfortran
 
     runs-on: ${{ matrix.os }}
     steps:
--- a/.github/workflows/build-and-validate.yml
+++ b/.github/workflows/build-and-validate.yml
@@ -22,7 +22,8 @@ jobs:
     - name: Install requirements
       run: ${{ matrix.install-script }}
     - name: Extra config
-      run: printf '%s\n' '${{ matrix.extra-config }}' >> config/configure_options/default
+      run: printf '%s\n' '${{ matrix.extra-config }}' >> config/configure_options/default \
+           && ( export LC_ALL=C; sort -o config/configure_options/default config/configure_options/default )
     - name: Build
       run: ./non_interactive_autogen.sh -s
     - name: Validate
--- a/.github/workflows/build-and-validate.yml
+++ b/.github/workflows/build-and-validate.yml
@@ -34,5 +34,5 @@ jobs:
       if: ${{ always() }}
       uses: actions/upload-artifact@v2
       with:
-        name: Validation log file
+        name: ${{ matrix.os }} validation log
         path: ./validation.log
--- a/.github/workflows/build-and-validate.yml
+++ b/.github/workflows/build-and-validate.yml
@@ -4,9 +4,10 @@ on: push
 
 jobs:
   build:
-
-    runs-on: ubuntu-latest
-
+    strategy:
+      matrix:
+        os: ['ubuntu-latest', 'macos-latest']
+    runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v2
     - name: Install requirements
--- a/.github/workflows/build-and-validate.yml
+++ b/.github/workflows/build-and-validate.yml
@@ -4,13 +4,20 @@ on: push
 
 jobs:
   build:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
+            install-script: sudo apt-get install
+          - os: macos-latest
+            install-script: brew install
 
-    runs-on: ubuntu-latest
-
+    runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v2
     - name: Install requirements
-      run: sudo apt-get install make automake libtool libtool-bin autoconf doxygen gfortran g++ 
+      run: ${{ matrix.install-script }} make automake libtool libtool-bin autoconf doxygen gfortran g++ 
     - name: Build
       run: ./non_interactive_autogen.sh -s
     - name: Validate
--- a/.github/workflows/build-and-validate.yml
+++ b/.github/workflows/build-and-validate.yml
@@ -14,7 +14,9 @@ jobs:
             install-script: |
               brew install make automake libtool autoconf doxygen
               brew reinstall gfortran
-            extra-config: --enable-suppress-cgal-build
+            extra-config: |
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by PuneetMatharu** [comment]
Great work. Thanks for sorting that out. I have three comments/notes on this commit:

1. It would be helpful to add support/testing for MacOS. Could you add it to the platforms to run the tests on?
2. I've found it time-consuming to load the large amount of output from the GitHub Actions page. It would be nice to have a copy of the output instead. Realistically, we would only require the `validation.log` file to find/diagnose issues with failing tests, so could you upload this file from within the YAML workflow file? For example, you could add the following 
    ```yaml
          - name: Upload log file
            uses: actions/upload-artifact@v2
            with:
              name: Validation log file
              path: ./validation.log
    ``` 
    If you have the time, it would also be nice (but not necessary) to run the `script` command and upload the output file. I think this is reasonable since we're only running tests on Unix-based systems, where the command is available.
3. I have encountered this failing demo_driver before. IIRC, it will fail on a single number in the `validata`. I propose one of two solutions be used to fix this issue, either: (i) update the "largest relative error" value used by `fpdiff.py` in the `validate.sh` script, or (ii) update the numerical zero value. (I would probably prefer the latter of the two.) Values of `0.2` and `2.0e-08` would fix these issues on my end. However, before we commit to specific values, I would like to see what values appear in the `validation.log` file to make sure that this would fix the issue for the CI workflow. For this, I refer to point 1 above.

**Review by PuneetMatharu — CHANGES_REQUESTED** [review]
Great work. Thanks for sorting that out. I have three comments/notes on this commit:

1. It would be helpful to add support/testing for MacOS. Could you add it to the platforms to run the tests on?
2. I've found it time-consuming to load the large amount of output from the GitHub Actions page. It would be nice to have a copy of the output instead. Realistically, we would only require the `validation.log` file to find/diagnose issues with failing tests, so could you upload this file from within the YAML workflow file? For example, you could add the following 
    ```yaml
          - name: Upload log file
            uses: actions/upload-artifact@v2
            with:
              name: Validation log file
              path: ./validation.log
    ``` 
    If you have the time, it would also be nice (but not necessary) to run the `script` command and upload the output file. I think this is reasonable since we're only running tests on Unix-based systems, where the command is available.
3. I have encountered this failing demo_driver before. IIRC, it will fail on a single number in the `validata`. I propose one of two solutions be used to fix this issue, either: (i) update the "largest relative error" value used by `fpdiff.py` in the `validate.sh` script, or (ii) update the numerical zero value. (I would probably prefer the latter of the two.) Values of `0.2` and `2.0e-08` would fix these issues on my end. However, before we commit to specific values, I would like to see what values appear in the `validation.log` file to make sure that this would fix the issue for the CI workflow. For this, I refer to point 1 above.

**Comment by jondea** [comment]
1. It can be done, but I need a command to install the prerequisites on MacOS (tried the super naive approach, forgot that Mac doesn't use apt-get https://github.com/jondea/oomph-lib/actions/runs/1002652101)
2. Good idea, done

**Review by PuneetMatharu — CHANGES_REQUESTED** [review]
1. I think Homebrew is the go-to package manager for MacOS-based runner environments. Have you tried using `brew install` in place of `[sudo] apt-get`? All the packages you installed _except_ `g++` should be available via `brew`. However, in this case, I think the `g++` command should be automatically available (but is just an alias for the Clang compiler).
2. Fantastic, thanks for sorting that out. However, I did fail to recognise that in the event of a failure, the workflow will exit and fail to upload the required `validation.log`. To get around this, could you give the following a spin:
```yaml
    - name: Validate
      id: make_check
      run: make check -k -s
      continue-on-error: true

    - name: Upload validation log file on failure
      if: steps.make_check.outcome == 'failure'
      uses: actions/upload-artifact@v2
      with:
        name: Validation log file
        path: ./validation.log

    - name: Propagate status
      if: steps.make_check.outcome == 'failure'
      run: exit 1
```
3. Once point 2 has been addressed, can you provide a link to the output `validation.log`? So that we can see which new values are required.

Thanks!

jondea:
1. Thanks, I've re-added MacOS with brew, hopefully this time it should work!
2. Very good point! I took a slightly different approach but it should do the trick, link to build: https://github.com/jondea/oomph-lib/actions/runs/1006094721

jondea:
MacOS is failing, and I'm not sure how to debug https://github.com/jondea/oomph-lib/runs/3003896549?check_suite_focus=true

MatthiasHeilManchester:
Looks like it doesn't have a fortran compiler. That would be a problem. The configure options don't specify one. If you know what it's called on macos we could just specify ih with F77=...

PuneetMatharu:
Will take a look at this today. [Apparently `brew install gfortran` may not install `gfortran`.](https://github.com/qpv-research-group/solcore5/pull/161). Instead, we may need to use `brew reinstall gfortran`. I'll try to fork your repository @jondea today and see if making those changes works.

PuneetMatharu:
It appears that the linked GitHub issue was correct. I created a minimal test problem/repository to test this out. In one commit, I try to install the Fortran compiler using just `brew install gfortran`, and with the other, I user `brew reinstall gfortran`. The former fails and the latter passes. Links to Workflow Actions below:

- [Test with `brew install gfortran`](https://github.com/PuneetMatharu/check-macos-fortran-compiler-in-github-action/actions/runs/1007706586)
- [Test with `brew reinstall gfortran`](https://github.com/PuneetMatharu/check-macos-fortran-compiler-in-github-action/runs/3008287053?check_suite_focus=true)
**Note:** the first test fails on Ubuntu because of Fortran formatting issues; this was fixed for the second.

**TL;DR:** @jondea, can you update the `install-script` key please to do this? It should just be:
```yaml
- os: ubuntu-latest
  install-script: sudo apt-get install make automake libtool libtool-bin autoconf doxygen gfortran g++
- os: macos-latest
  install-script: |
    brew install make automake libtool autoconf doxygen
    brew reinstall gfortran
```

jondea:
Thanks for investigating that, done!

Just a note, if it's easier, you can push to this branch on my repo because I have set the "Allow edits and access to secrets by maintainers" to true on the PR

PuneetMatharu:
Excellent. Looks like we're almost there! 

Looks like there's an error (which I have painfully encountered many times) about an implicit function declaration in an ancient external source (the METIS which ships with ParMETIS). This is caught by a compiler enforcing C99 standards. Looking at my old configure options, I have a ` -Wno-implicit-function-declaration` flag to silence the warning which, rather annoyingly, manifests as an error). I haven't looked at how you've set up the `non_interactive_autogen.sh` (if it needed any extra tweaks) for the Workflow Action, but I imagine you might be the best person to address this. Would you be okay handling this? (I have a little too much on my plate at the moment otherwise I would have taken care of it personally.) The solution should be as simple as checking that the output of the `uname` command returns `Darwin` and, if true, append ` -Wno-implicit-function-declaration` to the `CFLAGS` variable.

MatthiasHeilManchester:
Have discussed this with Puneet and we decided the easiest way to get this checked in is for you to add the relevant lines
(below) into the relevant Makefile.am. As far as I can tell (on my non-mac!) the following should get the relevant flag into the compilation of the "metis from parmetis" sources.

So, addding these three lines to 

      external_src/oomph_metis_from_parmetis_3.1.1/Makefile.am

if HAVE_DARWIN
CFLAGS += -Wno-implicit-function-declaration
endif

should do the trick. (In my code I've put them under the definitions of LDFLAGS, i.e. below line 54:

mheil@mheil-Precision-7730:~/version_from_forked_github/oomph-lib/external_src/oomph_metis_from_parmetis_3.1.1$ git diff Makefile.am 
diff --git a/external_src/oomph_metis_from_parmetis_3.1.1/Makefile.am b/external_src/oomph_metis_from_parmetis_3.1.1/Makefile.am
index 8af1172..be6e0b4 100644
--- a/external_src/oomph_metis_from_parmetis_3.1.1/Makefile.am
+++ b/external_src/oomph_metis_from_parmetis_3.1.1/Makefile.am
@@ -53,6 +53,14 @@ lib_LTLIBRARIES = liboomph_metis_from_parmetis_3.1.1.la
 liboomph_metis_from_parmetis_3_1_1_la_SOURCES = $(headers_and_sources) $(extra_headers)
 liboomph_metis_from_parmetis_3_1_1_la_LDFLAGS = -static
 
+# Deal with implicit function declaration which causes problems on MacOS
+if HAVE_DARWIN
+CFLAGS += -Wno-implicit-function-declaration
+endif
+
+junk:
+       echo $(CFLAGS)
+
 # The library's include headers:
 #-------------------------------
 # Headers that are to be included in the $(includedir) directory:

MatthiasHeilManchester:
Update: the changes are also included in the pull request that I've just submitted.

PuneetMatharu:
That reminds me, for future reference, you can create, share and apply patches from Git commits (as described in the well-known Git book).

**Steps:**

1. Make your changes to the relevant files, then stage and commit the changes (i.e. `git add` then `git commit`) .
2. Get the ID of the commit just _before_ you added your change. For example, with my Git log below:

    ```bash
    >> git log -n 2 --oneline
    26330c63 (HEAD -> main, origin/main) My patched changes.
    19d130a6 Commit with breaking changes that need patching.   # <---- get the commit ID from this one
    ```
    I would then create a patch file using 
    ```bash
    git format-patch 19d130a6
    ```
    which produces something like
    ```bash
    0001-My-patched-changes.patch
    ```
3. You can send this to someone for use. For example, you could attach it here to a pull request message (just drag and drop). To apply it, use
    ```bash
    git am < 0001-My-patched-changes.patch
    ```
    which would apply the changes then commit them.

jondea:
> Update: the changes are also included in the pull request that I've just submitted.

Thanks, I rebased this branch onto main so I should have these changes now.

PuneetMatharu:
Just checked the status of your MacOS runner and the same problem appears to crop up in `oomph_superlu_*`, so we might need to rinse and repeat. I've got things tying me up for the next few days off, so I can't do it myself at the moment, unfortunately. I think we have two options:

1. Apply it globally (which should fix the issue), or
2. Apply it on a case-by-case basis. 

Given that we're only ignoring this warning in the compilation of C programs, disabling this warning should really only affect the external sources and not our libraries, so it might not be the worst thing in the world to do. Alternatively, if you're happy to park this for possibly a week(?), I can come back and address the warnings/errors individually.

MatthiasHeilManchester:
Still not sure I understand 100% how this works. At the moment nobody appears to have merged this in which probably means that we can't re-run the github-based self-tests (ubuntu and macos) to see if it works. Should I merge the pull request or are we waiting for anything else?

jondea:
I should probably rebase this branch with the fixes (or merge main into this). Then see if the runners pass. I can't do it on my phone, but I should be able to do it later. Have we fixed the clamped shell validation failure? Or are we going to do that after we have merged?

PuneetMatharu:
@MatthiasHeilManchester: The Workflow Actions are executed each time Jon pushes changes to his repo, so we only need to monitor the status of the tests under the Actions tab of his repo [here](https://github.com/jondea/oomph-lib/actions) to see whether it has fixed the tests. 

@jondea: The plan is to get the self-tests running on both Ubuntu and MacOS. This will allow us to identify `fpdiff.py` parameters for the `clamped_shell` demo driver that makes it pass on both platforms (rather than having to make multiple commits to tweak it for each OS).

PuneetMatharu:
Looks like the self-tests broke on the build of CGAL whilst building Boost. I encountered this issue in the past but just disabled the build of CGAL. As there's likely no know fix right now, I think it might be best to just disable its build for now so that the self-tests at least run on MacOS. I think there's a simple flag to do this in `configure.ac` along the lines of '`--enable-suppress-CGAL-build`. Any objections @jondea, @MatthiasHeilManchester? If so, is there a way to disable this by specifying it as an argument to `non_interactive_autogen.sh` so that it can be supplied from a Workflow script? (As we don't want to hardcode it within `configure.ac`.)

PuneetMatharu:
Great thinking on incorporating that addition. 

Quick note: the letters "CGAL" should also be in lowercase letters. (See line 923 of `configure.ac`.)

PuneetMatharu:
The build finishes now, which is good, but it falls down during linking. This is likely related to the issue with MacOS with not coping well with shared libs. To fix this, I think you just need to supply the `--disable-shared` flag. Hopefully this is the last step to making it work on MacOS.

PuneetMatharu:
Great. Looks like the tests run as expected: `shell/plate` and `one_d_act_inhibit` both fail. The former will be because it exceeds the fpdiff tolerances, and the latter because of some unknown bug (which makes no sense) that leads to a seg. fault. You can ignore it.
 
As I suspected earlier, multiple workflows that upload a file with the same name appear to overwrite each other. You probably want to construct a unique ID for the file name based on the matrix parameters. Happy to merge it in as soon as that’s done.

PuneetMatharu:
Looking at the latest results with Jon's commits, there are two tests that fail:

1. `shell/clamped_shell` on Ubuntu (exceeds relative error tolerance; largest rel. error = 0.100097, tol = 0.1), and
2. `reaction_diffusion/one_d_act_inhibit` on macOS (mysterious seg. fault; reason unknown).

In a previous run with Jon's machinery, the `shell/plate` self-test also failed on macOS, due to a problem with `unstructured_clamped_square_plate.cc`. This failure produced warnings/an error like that mentioned in https://github.com/oomph-lib/oomph-lib/issues/12. However, it is likely that we are going to remove these tests because they are no longer needed (see discussion in linked issue). 

For the above reasons, I suggest that we squash and accept this PR as is (as it only affects a single file) and address the tolerances issues in a separate commit. Would you agree, @alhazel and @MatthiasHeilManchester?

jondea:
Given how simple the fix is (just removing the offending failing tests) it's probably worth merging the fix for #12 first so that we can verify that the pipelines do actually pass before merging them in. Agree with the squash though, given that it is a single file in the end.

PuneetMatharu:
Got it, that sounds good to me.

alhazel:
Agree with putting in the fix for #12 first. Odd that 1d_act_inhibit fails, there's not much to that code.

PuneetMatharu:
The `one_d_act_inhibit` driver is a bizarre one. I've been in there with a debugger before (a long time ago) and it just mysteriously dies jumping from constructor to constructor (of the elements). I might try to add sanitiser support to the CMake build in the future to see if I can get to the bottom of it.

PuneetMatharu:
@jondea, the problematic driver codes have been committed (see https://github.com/oomph-lib/oomph-lib/pull/13). Can you go ahead and rebase? I'll watch the self-test results and see if they all pass. Thanks!

jondea:
[Ubuntu passed but MacOS didn't](https://github.com/jondea/oomph-lib/actions/runs/1040497311), it's the one_d_act_inhibit again. What's the consensus with that? Are we ignoring it for now?

PuneetMatharu:
I think it's best to ignore it for now. However, it might become slightly annoying if we constantly get emails about the test failing upon every PR. It would also make it more painful to diagnose when a test on macOS fails (i.e. we would *always* have to check the `validation.log` to make sure that it is the `one_d_act_inhibit` driver that fails. It might be best to disable it on macOS.

As a workaround, we could update the `Makefile.am` so that if `HAVE_DARWIN` has the value true, we add `one_d_act_inhibit` to `noinst_PROGRAMS` instead of `check_PROGRAMS`. This should remove it from the self-tests but still allow someone sitting in the directory to build the driver. Would this be a sensible idea, @MatthiasHeilManchester, @alhazel?

MatthiasHeilManchester:
Agreed -- that way it's also transparent that it's a mac specific problem.

PuneetMatharu:
Turns out this won't work - I should have known better. The `one_d_act_inhibit` target gets built during the autogen process. When the self-tests are run, the executable has already been compiled and running `make check` afterwards calls the validate wrapper script which in turn executes the `validate.sh` script. 

To get around this issue without disabling the ability for a Mac user to compile the program, I think we need to intercept the `make check` command in `validate.sh`. I have a fix for this that simply returns an exit code of 0 (indicating a pass) inside `validate.sh` when `uname` yields the value `Darwin`. I've created a new PR (https://github.com/oomph-lib/oomph-lib/pull/14) to integrate this change.

PuneetMatharu (approved):

PuneetMatharu:
Congrats on what is likely to be the most painful PR in history!

PR #2, opened 2021-07-05
Update gitignore

jondea:
Various updates to `.gitignore` because it has become a little out of date since 2016


PuneetMatharu (changes requested):
Thanks for the additions! I think it may be worth removing the hardcoded binary/executable paths and using a glob recipe of the form:
```
# Ignore all
*
# Unignore all with extensions
!*.*
# Unignore all dirs
!*/
```
which should be placed at the top of the `.gitignore` file, before anything else. Additionally, glob expressions that are not preceded with a forward slash will be ignored in **all** subdirectories. They are only ignored in a single, specific directory if the expression begins with a forward slash. More info [here](https://www.atlassian.com/git/tutorials/saving-changes/gitignore). I think exploiting this would be helpful for things like `Validation` directories and `doc` output. For example, if all `.html` files, `extra_latex_style_files` files, and `Validation` directories should be ignored, you can simply use the following three lines:
```
Validation/
*.html
extra_latex_style_files
```
If the latter two rules should only be applied to the `doc` directory, you could place a `.gitignore` file with the latter two lines in the `doc` directory or use the following:
```
doc/**/*.html
doc/**/extra_latex_style_files
```
[Here is the gitignore.txt I've been using.](https://github.com/oomph-lib/oomph-lib/files/6766258/gitignore.txt) I hadn't got around to pushing it yet but I also think it may be overkill. Your version seems more appropriate so I would prefer to go off that. Could you try applying some of those simplifications to the `.gitignore` file when you get the chance? 

Thanks!

jondea:
Yeah, that's a much better approach, I think you have given this more thought than me. Probably better to reject this PR, would you be able to submit one with your changes @PuneetMatharu ?

PuneetMatharu:
Sounds good to me. I'll try to get around to it sometime today.

PR #3, opened 2021-07-06
Updated .gitignore file

PuneetMatharu:
As per #2, I've tidied up and condensed the `.gitignore` file by making use of the available glob recipes.


PuneetMatharu:
Accidentally deleted the branch from my fork before merging. Reopening pull request.

.gitignore
```diff
@@ -1,17 +1,35 @@
 ######################################################################
 # Below is mostly taken from the web. Common things to ignore for C++
 ######################################################################
+# Ignore all files then un-ignore those that have an extension; this has the
+# effect of ignoring all executables.
+#        **THESE THREE LINES MUST STAY AT THE TOP OF THE FILE.**
+*
+!*.*
+!*/
```
jondea, line 9: These three lines are non-trivial, might be worth putting a comment above each like before

.gitignore
```diff
@@ -39,31 +63,30 @@ make_check_output
...
-test-suite.log
-validate.sh.log
-validate.sh.trs
-test-driver
-self_test/analyse_self_tests/analyse.sh.log
-self_test/analyse_self_tests/analyse.sh.trs
-
+doc/preconditioners/lgr_navier_stokes/lgr_navier_stokes_old.txt
+doc/preconditioners/lgr_navier_stokes/lgr_navier_stokes_old2.txt
+doc/preconditioners/lsc_navier_stokes/.lsc_navier_stokes.txt.swo
```
jondea, line 86: You can remove this line and ignore all *.swo files. They are a temporary file produced by vim.

.gitignore
```diff
@@ -39,31 +63,30 @@ make_check_output
...
 
-########################################
-# Oomph-lib specific generated files
-########################################
+# Temporary files
+*.tmp
```
jondea, line 89: Merge with Temporary files higher up

.gitignore
```diff
@@ -133,461 +150,42 @@ external_src/oomph_arpack/debug.h
...
-demo_drivers/linear_solvers/two_d_poisson_cg
-demo_drivers/linear_solvers/two_d_poisson_dense_lu
-demo_drivers/linear_wave/two_d_linear_wave/two_d_linear_wave
-demo_drivers/linear_wave/two_d_linear_wave/two_d_linear_wave_flux
-demo_drivers/linear_wave/two_d_linear_wave_adapt/two_d_linear_wave_adapt
-demo_drivers/linking/demo_code
+bin/minimal_mpi_test.out
+
+demo_drivers/heat_transfer_and_melting/OUTPUT
+demo_drivers/heat_transfer_and_melting/result_spring_contact_kuhn_tucker.dat
```
jondea, line 167: Can we add a rule for *.dat and remove this (and all other dats?)

.gitignore
```diff
@@ -133,461 +150,42 @@ external_src/oomph_arpack/debug.h
...
-demo_drivers/linear_solvers/direct_solver_test
-demo_drivers/linear_solvers/driven_cavity
-demo_drivers/linear_solvers/two_d_linear_elasticity_with_simple_block_diagonal_preconditioner
-demo_drivers/linear_solvers/two_d_poisson_cg
-demo_drivers/linear_solvers/two_d_poisson_dense_lu
-demo_drivers/linear_wave/two_d_linear_wave/two_d_linear_wave
-demo_drivers/linear_wave/two_d_linear_wave/two_d_linear_wave_flux
-demo_drivers/linear_wave/two_d_linear_wave_adapt/two_d_linear_wave_adapt
-demo_drivers/linking/demo_code
+bin/minimal_mpi_test.out
```
jondea, line 164: *.out is already ignored, this may be redundant?

jondea (commented):
Looks really good, thanks, a few minor potential improvements

.gitignore
PuneetMatharu: Good idea! I'll take care of that.

.gitignore
PuneetMatharu: Ah, I see, thanks for letting me know.

.gitignore
PuneetMatharu: Nice observation, will take care of that now.

.gitignore
PuneetMatharu: Excellent catch, it is indeed. Will remove it now.

.gitignore
PuneetMatharu: Unfortunately not, there are a lot of raw (unpacked) `.dat` files in various folders (e.g. some `.dat` files just contain the number of iterations for convergence of a self-test using an iterative solver and do not require compressing). Ignoring them would break several self-tests. For this reason, I'm going to leaving this rule untouched.

PuneetMatharu (commented):
I've addressed the aforementioned snippets. Happy with the changes?

MatthiasHeilManchester:
Just a quick comment: My understanding is that files (and patterns) are ignored in the sense that they don't clutter up the screen when doing things like git status. If so there are a fair number of files that probably shouldn't be there. I'm happy to have another look through this at some point to tidy this up but happy for this to be committed now.

.gitignore
jondea: Ignoring them wouldn't remove already committed files or ignore changes, but I get what you're saying.

jondea (approved):

jondea:
> Just a quick comment: My understanding is that files (and patterns) are ignored in the sense that they don't clutter up the screen when doing things like git status. If so there are a fair number of files that probably shouldn't be there. I'm happy to have another look through this at some point to tidy this up but happy for this to be committed now.

That's one way to see it. It's anything that could be in your repo which you would never reasonably want to commit (or at least strongly discourage from committing). Build artifacts are the obvious ones, but metadata and temporary files are a good idea to have in there too.

.gitignore
PuneetMatharu: Yep, but it would make future additions more difficult. It would force the user to use `git add --force` to get their files in; if the user is using a glob expression for the staging then this is likely to cause extra things, that shouldn't be added, to be added in.

PR #6, opened 2021-07-08
Replaced licence boiler plate for all cc and h files in our sources

MatthiasHeilManchester:
(Basically removed svn revision numbers and updated the copyright date
to end in 2021, a.k.a. now).

Can somebody have a look to see if I'm doing the right thing? I'd like these changes to be pulled back into the main repository.

### DIFF
--- a/bin/cc_and_h_licence_block.txt
+++ b/bin/cc_and_h_licence_block.txt
@@ -3,11 +3,7 @@
 //LIC// multi-physics finite-element library, available 
 //LIC// at http://www.oomph-lib.org.
 //LIC// 
-//LIC//    Version 1.0; svn revision $LastChangedRevision$
-//LIC//
-//LIC// $LastChangedDate$
-//LIC// 
-//LIC// Copyright (C) 2006-2016 Matthias Heil and Andrew Hazel
+//LIC// Copyright (C) 2006-2021 Matthias Heil and Andrew Hazel
 //LIC// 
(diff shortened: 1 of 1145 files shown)


PuneetMatharu:
That looks great to me!

I think it is the right thing to set the copyright notice end year to the current year, but we might want to make sure to update this on a yearly basis. If you used a simple glob command, could you store it in a bash script and commit it to your repository on the same branch as this commit (i.e. `main`)? It should get added to this pull request. When I get time, I can create a Workflow Action that executes like a cronjob and updates the copyright date at the start of each year.

Unrelated to these changes, do you think you can update the `private/` files too before you upload them to a repository? 

Of course, if you're too tied up to add a bash script, I'm happy to sign these changes off as is, and I'll add it as a "to-do" to the project board. @jondea, what do you think? Happy too?

MatthiasHeilManchester:
There are two main scripts:

  bin/remove_licence.sh
  bin/apply_licence.sh

which do what they say (and use an awk script and the file bin/cc_and_h_licence_block.txt which contains the new licence text). 
Did it in two stages because I was scared when I first did it... Could merge them into one if that's what you wanted.

   M.

PuneetMatharu:
Excellent, I'm happy to call it a day and merge the changes in then. Also, no need to merge them, they'll be completely fine as is.

PuneetMatharu:
Congrats on your first pull request!

PR #8, opened 2021-07-09
Patch: fix implicit function declaration error on macos

PuneetMatharu:
Patch for https://github.com/oomph-lib/oomph-lib/pull/1. I've addressed the implicit function declaration warnings which present as errors on MacOS; I just repeated the strategy used by @MatthiasHeilManchester to update `oomph_lapack` and `oomph_superlu_4.3`.

Edit: don't approve this yet. It seems to have included several commits which have already been merged. Let me sort that out first.

### DIFF
--- a/external_src/oomph_lapack/Makefile.am
+++ b/external_src/oomph_lapack/Makefile.am
@@ -32,6 +32,13 @@ lib_LTLIBRARIES = liboomph_lapack.la
 liboomph_lapack_la_SOURCES = $(headers_and_sources)
 liboomph_lapack_la_LDFLAGS = -static
 
+# Deal with implicit function declaration which causes problems on MacOS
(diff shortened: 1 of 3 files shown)


MatthiasHeilManchester:
We're all too keen! I've just fixed this globally by adding the flag in configure.ac. As far as I can tell the problem also arises when compiling trilinos! It's a temporary thing anyway so I don't mind it being used everywhere until your cmake machinery takes over.
Im ready to submit a pull request (which also includes the deletion of the four accidentally checked in generated combined header files in your space_time directory.  If your commit isn't quite ready yet, shall we go with mine?

PuneetMatharu:
Just fixed mine, but it wouldn't address issues with Trilinos, so nice find on that end! Let's go with yours.

PR #10, opened 2021-07-09
no implicit fcts and deleted header files

MatthiasHeilManchester:
Here's my version for dealing with no implicit functions problem on MacOS. I've also deleted the accidentally checked in generated combined header files. Puneet: Not sure if you've already committed your (more fine-grained) fix to the first problem. If not please check and accept. If not let me know.


PuneetMatharu:
Excellent, I'll squash and merge. 

I can see that there are residual commits from the previous "licence boilerplate..." stuff and the merge. It is worth trying to rebase off an upstream branch to get up-to-date changes. If there are no conflicts with your changes, this can avoid creating an extra merge commit, which might be nicer when we want to retain the commit history of a PR but we also don't want repeated commits.

PuneetMatharu:
**Update:** @MatthiasHeilManchester has fixed the issues with the implicit function declaration, see https://github.com/oomph-lib/oomph-lib/pull/10. Can you try to rebase again to see if it finally fixes the issue? 

I can possibly foresee us encountering one more issue with CGAL not working (I've had to disable it in the past on my Mac), but it may just work. Fingers crossed!

PR #11, opened 2021-07-14
Pull request for Andrew; black box newton solver now in separate file; .gitignore stuff

MatthiasHeilManchester:
Couple of straightforward changes to be merged back into the main repository. Partly to give Andrew something to play with...

**Puneet:** for some reason autom4te.cache/ is still being listed as untracked even though it is listed in .gitignore already. Don't understand. Other than that git status is now clean for me.


MatthiasHeilManchester:
It seems that this pull request includes more commits than I thought (presumably, as Puneet recently mentioned to me) more recent ones get added to any outstanding ones. Can somebody please process this. It was supposed to be a test for Andrew but I'm happy for Puneet to do it.

PuneetMatharu:
Just had a look through and they all look like pretty simple changes. Going to go ahead and squash and merge (I think it's appropriate in this case).

jondea:
@MatthiasHeilManchester you will probably find it easier to work on branches off your fork where each branch is a separate piece of work, rather than putting everything on main.

FWIW, I don't think PRs like this should be squashed. Each commit should be a self contained thing, something you might want to revert or cherry-pick in isolation. If it was done to tidy up the merge commits, I don't think it's necessary. When a branch has lots of trivial merge commits from main, I think they disappear when it is merged into main.

PuneetMatharu:
I agree that distinct commits shouldn't be squashed but half of the commits were merge commits to catch up with the upstream repository. I think it makes sense to avoid adding the superfluous/duplicated commits (or at least 3-4 of them) in a single small PR. Not to mention, it isn't terribly difficult to see what is done in the commit from looking at the affected files and the commit message (I purposely retained the individual commit messages for this purpose). 

In the future, the extra commits can probably be avoided by using `git pull --rebase`, which I've discussed with MH.

PR #16, opened 2021-07-22
Cleaned up most of the broken documentation where problems arose from…

MatthiasHeilManchester:
… missing doxygen hooks.

There's a remaining problem in navier_stokes/surface_theory/surface_theory.txt that Andrew will hopefully be able to sort out.

Also added CXXFLAGS to my own configure options file to allow c++11 style compilation (in anticipation of format changes to be introduced by clang-format).


PuneetMatharu:
Just as a note for anyone that comes across this PR through the commit log: the `run_clang_format_on_all.sh` applies the formatting specified in the `.clang-format` file. I noticed, however, when I was using it, that it doesn't actually ignore the `cfortran.h` and `lapack_qz.h` headers in `src/generic` as desired. If you have time, it might be nice to fix the `find` rule so that it does.

PuneetMatharu:
Merged in as discussed (separately).

MatthiasHeilManchester:
Sorted problem with specification of excluded header files in find command in #18

PR #20, opened 2021-07-28
Convert key files to Markdown (e.g. README), split workflows by OS and add status badges

PuneetMatharu:
**Key changes/additions:** 
- Converted `README`, `INSTALL`, `AUTHORS`, and `NEWS` to Markdown.
- Updated text in `README.md` and `INSTALL.md` to make use of Markdown formatting.
- Renamed `COPYING` to `LICENSE`.
- Split the `build-and-validate.yml` file into two separate files for each tested OS. 
   **Update (as recommended by Jon):** I split the workflow files into two separate OS-specific ones so I can add individual status badges for the self-tests on each OS.
- Updated `README.md`:
  - Added the famous `oomph-lib` logo to the `README.md`.
  - Added status badges to the `README.md` to display the license, the status of the tests, and the status of `build-and-publish-docs.yml`. 
- Renamed Workflow `.yml` files to have a `.yaml` extension.

**Remark:** The status badges appear broken in my repository (if you try to view them there) because I have updated the paths for the workflow files to reference the official repository and the renamed/split workflow files aren't in the official repository yet; I showed MH what the badges look like before changing the paths, so he knows they work.


INSTALL.md
```diff
@@ -0,0 +1,202 @@
...
+   `oomph-lib`s top-level directory contains a shell script [autogen.sh](autogen.sh)
+   which guides the user through the installation procedure and suggests
+   sensible defaults for various configure/installation options.
+   We highly recommend using this script to perform the installation.
+
+**2. Installation using the standard GNU build process**
+
+   You may, of course, also use GNU's standard
+
+      configure && make && make install && make check
```
jondea, line 18: Backticks

INSTALL.md
```diff
@@ -0,0 +1,202 @@
...
+  5. You can remove the program binaries and object files from the
+     source code directory by typing `make clean`.  To also remove the
+     files that `configure` created (so you can compile the package for
+     a different kind of computer), type `make distclean`.  There is
+     also a `make maintainer-clean` target, but that is intended mainly
+     for the package's developers.  If you use it, you may have to get
+     all sorts of other programs in order to regenerate files that came
+     with the distribution.
+
+Compilers and Options
```
jondea, line 74: Replace ===== with ##, we should probably be consistent, and this should probably be a h2 or h3 (===== means h1). There's a few more of these.

.github/workflows/self-tests-ubuntu.yaml
```diff
@@ -0,0 +1,31 @@
...
+        include:
+          - os: ubuntu-latest
+            install-script: sudo apt-get install make automake libtool libtool-bin autoconf doxygen gfortran g++
+
+    runs-on: ${{ matrix.os }}
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install requirements
+      run: ${{ matrix.install-script }}
+    - name: Extra config
```
jondea, line 19: For example, this stuff can be removed for the Ubuntu only workflow

INSTALL.md
```diff
@@ -0,0 +1,202 @@
...
+=====================
+
+   Some systems require unusual options for compilation or linking that
+the `configure` script does not know about.  You can give `configure`
+initial values for variables by setting them in the environment.  Using
+a Bourne-compatible shell, you can do that on the command line like
+this:
+
+     CC=c89 CFLAGS=-O2 LIBS=-lposix ./configure
```
jondea, line 83: Triple backticks

README.md
```diff
@@ -0,0 +1,24 @@
+[![oomph-lib](doc/figures/oomph_logo.png)](http://oomph-lib.maths.man.ac.uk)
+
+[![License: LGPL v2.1](https://img.shields.io/badge/License-LGPL%20v2.1-blue.svg)](https://www.gnu.org/licenses/lgpl-2.1)
+[![Ubuntu tests status](https://github.com/oomph-lib/oomph-lib/actions/workflows/self-tests-ubuntu.yaml/badge.svg)](https://github.com/oomph-lib/oomph-lib/actions/workflows/self-tests-ubuntu.yaml/badge.svg)
```
jondea, line 4: Ahh nice, thanks for adding badges! I now see why you wanted separate workflows, would have been useful to explain this in the PR body, it seemed like a strange choice initially.

INSTALL.md
jondea, line 83: Triple backticks

INSTALL.md
```diff
@@ -0,0 +1,202 @@
...
+
+Or on systems that have the `env` program, you can do it like this:
+
+     env CPPFLAGS=-I/usr/local/include LDFLAGS=-s ./configure
+
+Compiling For Multiple Architectures
+====================================
```
jondea, line 90: Consistent syntax for headers, this should probably be a h2 or a h3 too

jondea (changes requested):
Thanks for doing this, should make things a bit more polished! Just a few comments.

.github/workflows/self-tests-macos.yaml
```diff
@@ -1,4 +1,4 @@
-name: Build and validate
+name: macOS self-tests
```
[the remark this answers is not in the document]
PuneetMatharu: I initially thought the same about the matrix, but I think it might be useful to keep it in to make it easy for us if we want to add testing for different versions of the same OS (e.g. Ubuntu 18.04 as well as Ubuntu 20.04). 

Unfortunately, I don't think it's possible to avoid the code duplication here. AFAIK, the YAML language can't perform such logic; it has no concept of "import" or "include".

INSTALL.md
PuneetMatharu: An indented block *does* have the same effect, but I agree it's probably best to add them in. I'll take care of it. Thanks for the spot!

INSTALL.md
PuneetMatharu: *facepalm* 
Nice catch! I completely missed that whilst changing the rest of the text. I'll take care of it now. I also think `##` works best for this case so I'll stick to that (unless, of course, you'd prefer I go for h3).

.github/workflows/self-tests-ubuntu.yaml
PuneetMatharu: See https://github.com/oomph-lib/oomph-lib/pull/20#discussion_r678641150.

INSTALL.md
PuneetMatharu: Done.

README.md
PuneetMatharu: Good idea! I've updated the PR text accordingly.

INSTALL.md
PuneetMatharu: Done!

INSTALL.md
PuneetMatharu: Odd, this appears to be a duplicate. Addressed, as per https://github.com/oomph-lib/oomph-lib/pull/20#discussion_r678651643.

MatthiasHeilManchester:
Just a quick comment that I'm happy with all these but don't want to merge until Jonathan's happy with the stuff only he knows about.

Puneet: If you've discussed these with Jonathan and he's happy, let me know, then I'll merge it.

.github/workflows/self-tests-macos.yaml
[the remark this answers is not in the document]
PuneetMatharu: **Update:** Removed the matrix as mentioned.

.github/workflows/self-tests-ubuntu.yaml
PuneetMatharu: **Update:** Removed the abstraction with a matrix and tidied the script up.

PuneetMatharu:
Tests pass, as expected. Feel free to merge this in when ready, @MatthiasHeilManchester.

MatthiasHeilManchester:
Puneet told me it's ok to merge this!

PR #21, opened 2021-07-29
Add cronjob workflow to update copyright notice every year

PuneetMatharu:
**Description:**
- Automates the updating of the copyright notice headers on a yearly basis (I mentioned adding this earlier in https://github.com/oomph-lib/oomph-lib/pull/6).

**Key additions:**
- A shell script (`update_copyright_licence_header.sh`):

  - Updates the end-year in the copyright notice header (stored in `cc_and_h_licence_block.txt`) to the current year, then updates the headers of all `oomph-lib` sources.
  - Combines `apply_licence.sh` and `remove_licence.sh`, and adds functionality to change the end-year in the copyright notice line.

- A workflow script (`update-copyright-licence.yaml`):

  - Executes at 00:01 on January 1st each year.
  - Runs the shell script mentioned above on the `development` branch and commits the changes to that branch.

**Additional notes**
- The workflow will only add commits to the `development` branch.

  - I believe this is the right branch to update; if we add the changes to `main` instead, the changes will be overwritten when we merge the `development` branch in (as files on the `development` branch will have the old copyright year).
  - The `main` branch will receive the changes when we merge the changes from `development` into `main`.
  - We *could* also process and update the `main` branch but I'm not sure this is a great idea as this would create duplicate commits. However, this *would* only create duplicate commits once a year...
  - Users of the library should be able to just pull in/rebase on the added commit without issue as their branches should not have added changes to the copyright notice.
  - **I think it's worth discussing this point before merging the PR in.**

- All workflow scripts that operate on a schedule must be added to the default branch (`main` here) to work (hence why I haven't requested adding the changes to `development`).
- An unimportant point: a scheduled workflow will typically not run at the exact requested time (unlike triggered workflows), so don't expect it to. It will be placed in a queue and will be executed when resources are available.

**Minimal example:**
- I've created a minimal example [here](https://github.com/PuneetMatharu/ci-cronjob-example-update-copyright-notice). You can find the commit log [here](https://github.com/PuneetMatharu/ci-cronjob-example-update-copyright-notice/commits/main).
- To illustrate how the script works, the version associated with the "minimal example" updates the notice every hour today (29/07). Rather than setting the end-year in the notice to the current year (as this wouldn't change anything), it updates the end-year to the *current hour*.
- The workflow exists on the `main` branch but only updates the code in `src/` on the `development` branch, so switch to the commit log of the `development` branch to see the actual changes.

### DIFF
--- a/.github/workflows/update-copyright-licence.yaml
+++ b/.github/workflows/update-copyright-licence.yaml
@@ -1,0 +1,29 @@
+name: Copyright
+
+# IMPORTANT: A schedule workflow action *must* be placed on the default branch
+# of the repository. Only then will it run as scheduled. It is also placed in a
+# queue when scheduled, so it may not run exactly at the requested time.
+on:
+  # Calculated using crontab.guru. See here: https://crontab.guru/#1_0_1_1_*.
+  # Scheduled to run at 00:01 on January 1st of each year.
+  schedule:
+  - cron: '1 0 1 1 *'
+
+jobs:
+  update-copyright-notice:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Check out repository and switch to 'development' branch
+      uses: actions/checkout@v2
+      with:
+        ref: development
+
+    - name: Update copyright notice
+      run: ./bin/update_copyright_licence_header.sh
+
+    - name: Commit changes
+      uses: stefanzweifel/git-auto-commit-action@v4
+      with:
+        commit_user_name: copyright-notice-cron-bot
+        commit_message: 'Automated update of copyright notices.'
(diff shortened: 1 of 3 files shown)


PuneetMatharu:
I thought it was a pretty simple addition. If you'd rather/prefer to update the notices by hand every year, feel free to retire the issue.

PuneetMatharu:
**Note:** this PR will be updated in the near future by @MatthiasHeilManchester; do not merge it in yet.

PuneetMatharu:
**Update:** @MatthiasHeilManchester, I've gone ahead and shortened the bash script considerably to just edit the copyright notice rather than stripping the entire license header. I also deleted a duplicate license header in one of the source files.

Feel free to grab the bash script and run it on your own PC to test it out

PR #22, opened 2021-08-04
Namespace pollution

MatthiasHeilManchester:
This should be pretty straightforward...


MatthiasHeilManchester (commented):
Actually, having just had a look at this I don't understand what happened to these two files:

 src/meshes/quad_from_triangle_mesh.template.cc/.h

github seems to think that I've deleted them and then pasted everything back in...

PuneetMatharu:
Changes looks good and the self-tests pass so I'm happy to merge this in!

PR #24, opened 2021-08-07
Format sources with updated .clang-format and add clang-format workflow script

PuneetMatharu:
**Description:** 
- The first commit in the PR contains:
  - Updates to the `.clang-format` file, and;
  - A new workflow script that automates the formatting of the C++ code (with `clang-format`) in `src/` (except `cfortran.h`, `lapack_qz.h`, `superlu.c`, and `superlu_dist.c` in `src/generic/`) based on the `.clang-format` file.
- The second commit involves a switch to a personal fork of the `clang-format` repository.
- The third commit in this PR contains the updated sources, i.e. the sources after being formatted by the automated GitHub Action.

**Notes:** The workflow is only triggered in the following cases:
- After a pull request to the `development` branch; these changes will later be integrated into `main` when it is merged in.
- After a push to any branch *except `main` and `development`*, as these branches should never be updated directly by a user, and should only be updated through pull requests.


.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,41 @@
...
+    # Format all files in the src/ directory except cfortran.h, lapack_qz.h,
+    # superlu.c, superlu_dist.c, which sit in the src/generic/ subdirectory.
+    - uses: PuneetMatharu/clang-format-lint-action@v0.13
+      with:
+        files: ^(src)\/(?!.*(cfortran|lapack_qz))(.+)\.(h|cc)$
+        clangFormatVersion: 12
+        style: file
+        inplace: True
+
+    - uses: stefanzweifel/git-auto-commit-action@v4
```
MatthiasHeilManchester, line 38: Only glanced through this. This probably means that you're using an "action" defined by somebody else. What if this somebody else breaks/removes that action? Would it be safer to copy it across?

MatthiasHeilManchester (commented):
That looks fine to me. Two questions:

- Are we sure we want spaces between operators? The current setup actively changing things from 
 
       x+=1;
       x=sin(x)+3;

   to 

      x += 1;
     x = sin(x) + 3;

  It not how I'd do it (not that that matters so much) but I can't quite decide if the "spacing things out" is good or bad. Is it easy to change it the other way?

- Can we check if the reformatting breaks the documentation? In fact, this should definitely be added to the self-tests (whenever the documentation is built). There's a script that looks for suspicious patterns in the doxygen-generated html:

     bin/find_missing_doxygen_hooks.sh

.github/workflows/clang-format.yml
PuneetMatharu: Yes, it does. It's a well established repository so I would be very surprised if it breaks (although it is not impossible). The authors also use a semantic versioning system, so it should not break anything for us as long as we stick to a specific major version (which is why is what the `@4` is used to specify). _If_ it ever breaks, we can just add a hotfix to update the workflow to use the most recent non-breaking minor version. 

If you definitely want us to have our own forked copy, it's worth bearing in mind that our other workflow scripts also use actions that other people have published, so if we're going to copy things over, we might as well copy all of them over. It's worth running this by @jondea too if so.

PuneetMatharu:
> That looks fine to me. Two questions:
> 
> * Are we sure we want spaces between operators? The current setup actively changing things from
>   ```
>    x+=1;
>    x=sin(x)+3;
>   ```
>   to
>   ```
>   x += 1;
>   x = sin(x) + 3;
>   ```
>   It not how I'd do it (not that that matters so much) but I can't quite decide if the "spacing things out" is good or bad. Is it easy to change it the other way?
For the sake of readability, I think it is nicer. At the moment, there doesn't appear to be a way to remove/keep spaces around operators.
> * Can we check if the reformatting breaks the documentation? In fact, this should definitely be added to the self-tests (whenever the documentation is built). There's a script that looks for suspicious patterns in the doxygen-generated html:
>   bin/find_missing_doxygen_hooks.sh
Sure, as an aside or as part of the workflow? I can create a PR to add this to the `build-and-deploy-docs.yaml` script when I get time.

MatthiasHeilManchester:
OK, fair enough. How do I get this (uncommitted) branch

https://github.com/PuneetMatharu/oomph-lib/tree/feature/add-clang-format-workflow-script

onto my computer so I can build the documentation and check if any of the html gets broken before we trust the GitHub-based build machinery to check this for us?

PuneetMatharu:
In the green box (just above this text box), you should see the "Rebase and merge" button. Further to the right of this, you should see a link to "view command line instructions". Step 1 outlines how to obtain my changes, e.g.
```bash
git checkout -b PuneetMatharu-feature/add-clang-format-workflow-script development
git pull https://github.com/PuneetMatharu/oomph-lib.git feature/add-clang-format-workflow-script
```
After this branch has been checked out, you can perform any checks you want to.

PuneetMatharu:
Looking back at this, is there any chance you could give me more information on what I should expect from `bin/find_missing_doxygen_hooks.sh`? I can't just depend on the exit code of the script because it always exits normally. Would I need to analyse the output of the script? If so, it's worth noting already that on the current `gh-pages` branch, I get a non-zero number of "broken includes" matching "Pattern 2" (see below). At the moment, it is non-trivial to spot what could be broken by my `clang-format` changes.

#### Output from "checking for broken includes":
```bash
Checking for broken includes:
-----------------------------
 
Any files listed below contain html code
that looks suspiciously as if it's been generated from
a doxygen code include that went wrong -- usually because 
the hook in the c++ source code got changed and doxygen 
doesn't know what to include any more, resulting in an empty box. 
Output contains one previous line of text (from the generated html file)
to facilitate tracking down where the problem occurs.
 
Pattern 1: 
 
 
Pattern 2: 
 
./doc/navier_stokes/surface_theory/html/index.html-<div class="line">  }</div>
./doc/navier_stokes/surface_theory/html/index.html:</div><!-- fragment --><p>The additional contributions to the residuals and jacobian arise from the Lagrange multiplier contributions to the equations that determine the position of the nodes. The essential loop is the contribution below which adds the normal traction to the governing equations of solid mechanics: </p><div class="fragment"></div><!-- fragment --><h1><a class="anchor" id="surface"></a>
./doc/shell/linear_shell/html/index.html-<p>The namespace <code>Physical_Variables</code> is where the source function and the exact solution are defined. The source function can be specified via <code>Physical_Variables::source_function()</code> while the exact solutions are defined via <code>Physical_Variables::get_exact_u()</code>. Note that the six exact solutions correspond to the six degrees of freedom defined on each node. Furthermore, the applied source functions are required to be in the direction of the unit normal vector.</p>
./doc/shell/linear_shell/html/index.html: <div class="fragment"></div><!-- fragment --><hr  />
--
./doc/shell/linear_shell/html/index.html-<p>Following the usual self-test, we call the function <code>MyLinearisedShellProblem::parameter_study()</code> to compute the solution of the problem within a range of external pressures.</p>
./doc/shell/linear_shell/html/index.html: <div class="fragment"></div><!-- fragment --><hr  />
--
./doc/shell/linear_shell/html/index.html-<p>In the present problem, the function <code>Problem::actions_after_newton_solve()</code> is not required, so it remains empty. Also, the class includes two private data members which store pointers to a source function and to the geometric object that specifies the shell's undeformed shape.</p>
./doc/shell/linear_shell/html/index.html: <div class="fragment"></div><!-- fragment --><hr  />
--
./doc/shell/linear_shell/html/index.html-<p>The problem constructor starts by overloading the function <code>Problem::mesh_pt()</code> and set to the specific mesh used in the problem. In this tutorial, we implement the problem with 2D triangular unstructured mesh which is externally created by <code>Triangle</code>. The generated output will be used to build <code>oomph-lib</code> mesh. The reader may refer to <a href="../../../meshes/mesh_from_inline_triangle/html/index.html">another tutorial</a> to create an unstructured triangular mesh internally.</p>
./doc/shell/linear_shell/html/index.html: <div class="fragment"></div><!-- fragment --><p>We then create the undeformed centreline of the shell to one of an oomph-lib's standard shell geometric objects.</p>
./doc/shell/linear_shell/html/index.html:<div class="fragment"></div><!-- fragment --><p>Prior to consider the boundary conditions, we will illustrate how to take care extra nodes on the element for the <img class="formulaInl" alt="$ C^1 $" src="form_0.png"/>-interpolations. Since there is no degree of freedom of the <img class="formulaInl" alt="$ C^1 $" src="form_0.png"/>-interpolations defined on the mid-sided nodes, they have to be pinned.</p>
./doc/shell/linear_shell/html/index.html-<p>In order to do this, firstly, we loop over the element and pin all degrees of freedom on non-vertex nodes that associated with the <img class="formulaInl" alt="$ C^1 $" src="form_0.png"/>-interpolations. Since nodes in the element situate anticlockwise and the midside nodes fill in progressing along the consecutive edges, the pinning procedure is easily done by starting to pin from the third node and so on.</p>
./doc/shell/linear_shell/html/index.html:<div class="fragment"></div><!-- fragment --><p>Next, the boundary conditions of the problem will be taken care. We pin the nodal values on the boundaries where the boundary conditions applied. Note that, at the clamped boundaries, all second-order derivatives degrees of freedom are also pinned in order to reduce the number of degrees of freedom in the problem. These second-order derivatives are the derived boundary conditions that can be taken care by the natural boundary conditions.</p>
./doc/shell/linear_shell/html/index.html:<div class="fragment"></div><!-- fragment --><p>We then loop over the elements and set the pointer to the physical parameters (if any), the function pointer to the source function, and the pointer to the geometric object that specifies the undeformed surface of the shell.</p>
./doc/shell/linear_shell/html/index.html:<div class="fragment"></div><!-- fragment --><p>We finish the constructor by assigning the equation numbering scheme. </p><div class="fragment"></div><!-- fragment --><hr  />
--
./doc/shell/linear_shell/html/index.html-<p>In the <code>action_before_newton_solve()</code>, the problem specifications will be updated before performing the newton solve. The boundary values will be (re)set from the exact solutions.</p>
./doc/shell/linear_shell/html/index.html: <div class="fragment"></div><!-- fragment --><hr  />
./doc/deliberately_broken_doc_for_checking/html/index.html-<p>...and here's one that doesn't because I've deliberately broken the hook:</p>
./doc/deliberately_broken_doc_for_checking/html/index.html: <div class="fragment"></div><!-- fragment --><p> <br  />
 
Done checking for broken includes
```

MatthiasHeilManchester:
That (the output from the broken links checker) looks fine. The surface theory thing is on Andrew's todo list; the linear shell is probably gone (because we killed the associated driver codes, leaving the deliberately broken one. I'll double check on my machine (also to see where the problems with the alphabetical list and the odd title on the title page come from; I suspect it's a different version of doxygen).

MatthiasHeilManchester:
OK; that looks fine. Breaks a little of bit of documentation but I can tidy this up later. I suggest that I accept this pull request, then update my own version and then fix the problems (not sure what would happen if I fixed them in my local version of the branch I've checked out from you (using the instructions above) and then submitted a pull request -- seems like all hell could break lose, though probably not -- git is quite clever...).

   Anyway, given that this is a fairly substantial change to our code base, I really want to get this right and would like a quick chat before I do this. Can then also discuss what to do about the "git stash" thing.

PuneetMatharu:
Excellent, and I'd like to have a chat before the final sign-off too. I'll send you an email to arrange a meeting!

PuneetMatharu:
@MatthiasHeilManchester I've gone ahead and added information about `clang-format` to `CONTRIBUTING.md` (as we discussed) and I've just added the pre-commit machinery with relevant additions to the `CONTRIBUTING.md`. Let me know what you think.

As the changes to `CONTRIBUTING.md` are likely to conflict with changes made in https://github.com/oomph-lib/oomph-lib/pull/26, I'll add a comment to that PR to hold off until these changes have gone through.

PR #26, opened 2021-08-09
Final tweaks to contributing

MatthiasHeilManchester:
I think I may have submitted a previous pull request to main. If so, ignore; here's the real thing, including all the changes to the instructions and slight tweaks to wording. (I will kill the other pull request if I find it).


CONTRIBUTING.md
```diff
@@ -229,6 +229,12 @@ This involves the following steps:
 
    Provide a meaningful description of your changes in the textbox, press the button "Create pull request" and wait for somebody to merge your changes in (or get back to you with comments/requests). Note that subsequent changes (in response to discussions/requests, say) can simply be submitted to the same branch (repeating everything from step 3 above); they will automatically be added/included to the same pull request.
 
+   Note that when you return to the branch's GitHub page you won't see the lovely green ![](doc/README/compare_and_pull_request_button.png) button any more (presumably because a pull request is already pending). To add the newly made changes to the pull request, click on the "Contribute" button, next to the "Fetch upstream" one
```
PuneetMatharu, line 232: Can we remove the image embedded in the middle of the sentence? It looks rather messy at the moment. We could get away with just saying '...green "Compare & pull request" button...' instead of '...green \!\[\]\(doc/README/compare_and_pull_request_button.png\) button...'.

MatthiasHeilManchester: Done! (Lovely trial run through the updating a pending pull request -- I'm no longer scared!)

PuneetMatharu:
I've tidied up the discussion on the stashing/deleting changes so that readers are more aware of what each option does. Can you give it a quick read (when you have time, of course) to see if it sounds okay? When you click on the commit I just made to your branch, you should see three dots on the right where you can "View file", "Edit file", "Delete file". To the left of that is an icon that looks like a file that you can use to "Display the rich diff"; this is like `latexdiff`.

MatthiasHeilManchester:
Not so keen on the stash discussion. It needs work (e.g. will the changes disappear from the accidentally-changed-into branch once you return to the one you working on to do "git stash", etc.) and has an element of mission creep to it. The document is already quite long... So, given that the quickest way forward is to not do anything (i.e. bin your changes, assuming you haven't changed anything in the substance of what I wrote) let's do that... 

   Can I undo your commit? At all? From here (the webpage)?

PuneetMatharu:
Got it, but you should be aware that some details on stashing were already in there. However, they didn't make it clear how to use the stashed changes (thus making it plain unhelpful), so it's worth stripping those comments out too. Also, I don't think it was emphasised that the user will lose _**all**_ of their work by running `git reset --hard HEAD`, which is dangerous - they may have done a considerable amount of work and it would all simply be lost. At the very least, point (i) should be in there (i.e. you can just return to the original branch and commit your changes), and it should be made abundantly clear that their work will be deleted.

For the above reasons, I don't think it makes sense to undo the changes; it's better to overwrite the document with the desired changes. However, to answer your question, I don't think you can undo the commit from the GitHub website, but you can (apparently) do it from GitHub desktop. I only know how to do it from the command-line.

MatthiasHeilManchester:
Agreed. Ish. I wanted to warn people that switching branches moves changes across which I found (and still find) very unexpected. I said "do git --help stash" for details (and you do too where you decide mission creep is about to set in). Happy to make the newly added points clearer -- will explain what I'm confused by next time we chat; am working my way through the clangformat stuff at the moment.

PuneetMatharu:
Bril. As with the clang-format PR, I'm happy to discuss the details when we meet.

CONTRIBUTING.md
```diff
@@ -179,11 +179,25 @@ This involves the following steps:
...
-   git stash
-   ```
-   to stash away your changes; see `git help stash` for details on how to reapply the stashed changes.
+   You can switch between branches at any point (e.g. `git checkout main` will get you onto the `main` branch in your local repository) but if you have not committed your work, the changes will automatically be moved across to the branch you are switching to. This is unlikely to be the desired outcome as we are specifically working on a separate branch to keep our new work isolated from the rest of the code. If you accidentally ended up moving changed files onto another, otherwise clean branch, all is not lost! In this situation, you have three options: 
+     1. Switch back to the branch you were just on and commit your changes (`feature/add-new-important-headers` in this case). When you switch to the `main` branches, these changes will no longer follow you.
+     2. Save your changes for later by first running
+     
+        ```bash
+        git stash
+        ```
```
PuneetMatharu, line 188: Missing `>>>` in command.

CONTRIBUTING.md
```diff
@@ -179,11 +179,25 @@ This involves the following steps:
...
+     2. Save your changes for later by first running
+     
+        ```bash
+        git stash
+        ```        
+        and, once you have returned to the branch upon which you made changes, running
+        
+        ```bash
+        git stash apply
+        ```
```
PuneetMatharu, line 193: Missing `>>>` in command.

CONTRIBUTING.md
```diff
@@ -179,11 +179,25 @@ This involves the following steps:
...
+        
+        ```bash
+        git stash apply
+        ```        
+        If you ran `git stash` again before running `git stash apply`, you will need to identify which stashed changes to use; see `git help stash` for more details on how to do this.
+        Because this option is more likely to create issues when multiple stashes exists, we recommend that you only use it once you are familiar with Git.
+     3. Finally, the most aggressive option: get rid of these changes entirely. Use this option if you do not care about the changes you previously made. You can reset changes to all tracked files in your current branch since the last commit to that branch using
+        ```bash
+        git reset --hard HEAD
+        ```
```
PuneetMatharu, line 199: Missing `>>>` in command.

PuneetMatharu (commented):
Spotted a few instances of `>>>` missing from a code block. Thought I'd add a quick "review" to help point them out. There are also other code blocks that need to be changed in the same manner, but I can't highlight them in this review because you haven't edited them (see e.g. L41-42).

MatthiasHeilManchester:
Resolved conflicts within github (pain in the arse; web editor is very slow; do command line next time around).

PR #27, opened 2021-08-17
Explicitly demand C++11 support in build for macOS.

PuneetMatharu:
## Description of change

Explicitly demands C++11 support on macOS, fixes the issue with Makefile recipe in `simple_shear` demo-driver, and fixes the issue with the explicit instantiation of `SimpleRectangularQuadMesh` in `poisson/two_d_poisson/two_d_poisson2_mesh.cc`.

**Note:**  Unlike Ubuntu, the compiler on macOS does not appear to automatically assume C++11 as the standard, requiring the compiler flag to be specified explicitly.

## Affected components

- `bin/autogen_helpers.sh`: 
  - Added the function `InsertFlagsForEnablingC++11()` to sneak `-std=c++11` into the `CXXFLAGS` variable inside the user's configuration file to explicitly demand C++11 support.

- `non_interactive_autogen.sh`: 
  - Updated to call `InsertFlagsForEnablingC++11()` on macOS.

- `demo_drivers/solid/simple_shear/Makefile.am`: 
  - Updated custom Makefile recipe for `precompiled_mesh.o` to include the contents of `$(CXXFLAGS)` (to ensure that the C++11 support gets added on macOS).

  - **NOTE:** It appears that the aim of this recipe was to avoid adding optimisation (see lines below) without including `CXXFLAGS` but it might also go against the grain to explicitly provide the `-std=c++11` flag in there. 

  - Link to relevant lines: https://github.com/oomph-lib/oomph-lib/blob/43ade517cef185a8097591f8dcfbc7afa93710de/demo_drivers/solid/simple_shear/Makefile.am#L23-L27

  - **Q: Is it okay to keep `CXXFLAGS` in there?**

- `poisson/two_d_poisson/two_d_poisson2_mesh.cc`:
  - Updated to define the explicit instantiation *inside* the `oomph` namespace.

## Links to self-test workflows

- [**Ubuntu tests**](https://github.com/PuneetMatharu/oomph-lib/actions/runs/1138498968)
- [**macOS tests**](https://github.com/PuneetMatharu/oomph-lib/actions/runs/1138498971)

## Checklist

- [x] The Ubuntu tests pass.
- [x] The macOS tests pass.

PR #28, opened 2021-08-17
Add parallelisation to build and validation in self-test workflows

PuneetMatharu:
## Description of change

Adds parallel support to the build and validation steps of the self-test workflows.

**Remark:** The level of parallelisation is somewhat limited as Linux virtual machines have 2 cores and macOS virtual machines have 3 cores (details here: [GitHub-hosted runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners)).

## Affected components

- `.github/workflows/self-tests-ubuntu.yaml`: 
  - Now builds with 2 jobs and runs the self-tests using `bin/parallel_self_test.py`.

- `.github/workflows/self-tests-macos.yaml`: 
  - Now builds with 6 jobs and runs the self-tests using `bin/parallel_self_test.py`.

The number of jobs used for each OS was decided based on the results below.

## Links to self-test workflows

Given that Linux and macOS VMs only have 2 and 3 cores, respectively, I expected to max out the self-test speed with 2 jobs on Linux and 3 jobs on macOS. There was little variation in the timings when I used 2 jobs and 4 jobs on Ubuntu but, curiously, the workflow completed nearly 20% faster with 6 jobs than with 3 jobs on macOS. Results below:

- [Ubuntu (2 jobs):](https://github.com/PuneetMatharu/oomph-lib/actions/runs/1140091497): Time taken: 2h 2m.
- [Ubuntu (4 jobs):](https://github.com/PuneetMatharu/oomph-lib/actions/runs/1140091126) Time taken: 2h 4m.
- [macOS (3 jobs):](https://github.com/PuneetMatharu/oomph-lib/actions/runs/1140091500) Time taken: 1h 50m.
- [macOS (6 jobs):](https://github.com/PuneetMatharu/oomph-lib/actions/runs/1140091124) Time taken: 1h 31m.

## Dependencies

The PR builds on top of the patched changes in https://github.com/oomph-lib/oomph-lib/pull/27. To avoid the self-tests failing immediately, I recommend waiting until the PR linked is merged in before merging this one in. 

## Checklist

- [ ] The Ubuntu tests pass.
- [ ] The macOS tests pass.
- [ ] PR dependencies have been merged in (i.e. https://github.com/oomph-lib/oomph-lib/pull/27).

### DIFF
--- a/.github/workflows/self-tests-macos.yaml
+++ b/.github/workflows/self-tests-macos.yaml
@@ -20,7 +20,7 @@ jobs:
           && ( export LC_ALL=C; sort -o config/configure_options/default config/configure_options/default )
 
       - name: Build
-        run: ./non_interactive_autogen.sh -s -j$(nproc)
+        run: ./non_interactive_autogen.sh -s -j$(sysctl -n hw.logicalcpu)
 
       - name: Validate
         run: ./bin/parallel_self_test.py
--- a/.github/workflows/self-tests-macos.yaml
+++ b/.github/workflows/self-tests-macos.yaml
@@ -6,26 +6,27 @@ jobs:
   build:
     runs-on: macos-latest
     steps:
-    - name: Check out repository
-      uses: actions/checkout@v2
+      - name: Check out repository
+        uses: actions/checkout@v2
 
-    - name: Install requirements
-      run: |
-        brew install make automake libtool autoconf doxygen
-        brew reinstall gfortran
+      - name: Install requirements
+        run: |
+          brew install make automake libtool autoconf doxygen
+          brew reinstall gfortran
 
-    - name: Extra config
-      run: printf '%s\n' '--enable-suppress-cgal-build --disable-shared' >> config/configure_options/default \
-           && ( export LC_ALL=C; sort -o config/configure_options/default config/configure_options/default )
+      - name: Extra config
+        run:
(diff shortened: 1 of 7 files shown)


jondea:
You've said that 6 jobs is faster on macos, but your bullet point numbers say the opposite. 

Also, I'd be hesitant to put an actual number in there, might as well keep it general (and more self explanatory) and use `nproc` on ubuntu and something similar on macos.

You might find this interesting https://unix.stackexchange.com/questions/208568/how-to-determine-the-maximum-number-to-pass-to-make-j-option

It seems that two popular choices are nproc and nproc+1. I'd be happy to approve either option, either would be an improvement

PuneetMatharu:
Apologies, the timings were the wrong way around, but it would have been clear by following the links provided.

Good idea, although `nproc` isn't available on macOS. I've made the appropriate adjustments using `sysctl`. 

Thanks for this link, I wasn't sure what "`jobs`" should correspond to for the `make` command, e.g. the number of available threads or the number of available cores, which is why I performed the tests.

I've made the corrections and the self-tests are now re-running (linked below) with the updates. I still recommend waiting for the PR linked in the initial text to be merged in before accepting this one.
- [**Ubuntu**](https://github.com/PuneetMatharu/oomph-lib/actions/runs/1142906930)
- [**macOS**](https://github.com/PuneetMatharu/oomph-lib/actions/runs/1142906934)

PR #30, opened 2021-08-19
Tidy up the tutorials that broke because of clang-format

MatthiasHeilManchester:
What it says... There are two tutorials that still break: Andrew's free surface tutorial and a tutorial associated with the recently deleted unstructured shell element driver code. Am consulting with Andrew what to do about this so both of these are in his court.


PuneetMatharu:
I may not have understood the request the first time around so I'll ask: did you want to add Andrew's changes in this PR or a separate commit? (Just so I know whether this one is good to go!)

MatthiasHeilManchester:
Sorry; should have been clearer. I suppose strictly speaking it would be better if we got him to fix these as part of this PR but he's just been away for two weeks and isn't likely to get to this any time soon, so I'd rather get my fixes in there now.

PuneetMatharu:
Done!

PR #31, opened 2021-08-21
Make status badges point to the Actions page in the same repository.

PuneetMatharu:
## Description of change

Updates the links for the status badges in the `README.md` to link to, and display the status of, the tests in the user's own repository (when forked), rather than just those of the official repository.

## Affected components

- `README.md`: 

  -  I've updated the links in the status badges to: (i) point to the Actions page of the _same_ repository, not just the official one, and (ii) direct the user to the Actions page, rather than the badge link.

## Questions

Currently, the badges will monitor the status of tests on the default branch (which is `main`). As we typically only care about tests on the `development` branch (since `main` will always contain stable releases), **should I update the status badges to monitor the status of the tests on the `development` branch or both?**

For reference, [here is what I think would be the most helpful version](https://github.com/PuneetMatharu/oomph-lib/tree/scratch/test-html-for-status-badges).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 [![oomph-lib](doc/figures/oomph_logo.png)](http://oomph-lib.maths.man.ac.uk)
 
 [![License: LGPL v2.1](https://img.shields.io/badge/License-LGPL%20v2.1-blue.svg)](https://www.gnu.org/licenses/lgpl-2.1)
-[![Ubuntu tests status](https://github.com/oomph-lib/oomph-lib/actions/workflows/self-tests-ubuntu.yaml/badge.svg)](https://github.com/oomph-lib/oomph-lib/actions/workflows/self-tests-ubuntu.yaml/badge.svg)
-[![macOS tests status](https://github.com/oomph-lib/oomph-lib/actions/workflows/self-tests-macos.yaml/badge.svg)](https://github.com/oomph-lib/oomph-lib/actions/workflows/self-tests-macos.yaml/badge.svg)
-[![Documentation](https://github.com/oomph-lib/oomph-lib/actions/workflows/build-and-publish-docs.yaml/badge.svg)](https://github.com/oomph-lib/oomph-lib/actions/workflows/build-and-publish-docs.yaml)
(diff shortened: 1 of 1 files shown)


MatthiasHeilManchester:
Clever!

PuneetMatharu:
Apologies, I should have been clearer - this branch doesn't contain the changes to the `README.md` that I've linked. Those changes live on a `scratch` branch that I wanted to use to illustrate the changes. As I think _those_ were the changes you wanted, I'll add them in directly to the `development` branch.

MatthiasHeilManchester:
Ooops -- sorry, I overlooked the question (though my comment about it being clever still applies!). But yes, let's provide the status update tor the development branch. Is there any way to make it explicit that it refers to that?

PuneetMatharu:
Ah, I've provided the status badge for both branches now; take a look at the `README.md` on the `development` branch. Does that look reasonable?

MatthiasHeilManchester:
Yeah, that looks great! However, I've clicked on INSTALL.md and noticed that this duplicates information that I've just updated on the webpages. Duplication is bad (and I'm not totally sure if gnu build process actually works as described; have removed it from the main documentation because I wanted to keep things simple by only explaining the autogen.sh -- the less we explain the fewer mistakes we'll make!). Should we turn INSTALL.md into a very minimal file that mainly links to the webpage-based documentation?

PuneetMatharu:
Excellent.

I definitely agree (i.e. duplication is bad). The guide online does look slightly nicer to read. Do you think it would be possible to also provide a link to the relevant doc page after the documentation has been generated? (So that the user has an offline version to read.)

MatthiasHeilManchester:
Can/will do. Related to that: I've just realised that the webpages at     

      https://oomph-lib.github.io/oomph-lib/doc/html/index.html   

are not the most recent ones (following my changes). This is presumably because
they are generated off the main branch (which makes sense). If so, can we update the main 
branch now so people can see "the real thing"? I'm asking for extra information (e.g. oomph-lib
based publications, collaborators etc.) that I've partially updated.

PuneetMatharu:
Done! Might take some time for the web pages to get generated though

PR #35, opened 2021-08-25
Update change log and a few more tweaks for 2.0 release

MatthiasHeilManchester:
- Went through svn diff (from old repo) to check changes since last
  official release and tried to extract highlights. To encourage others
  to add to list in
    `https://oomph-lib.github.io/oomph-lib/doc/change_log/html/index.html`
- Stripped out most of the stuff in INSTALL.md and refer to webpage for
  installation instructions.
- Fixed typo in CONTRIBUTING.md

-


MatthiasHeilManchester:
If/when you sign this off can you also immediately push it to main, so people can look at the change log and add anything that's missing. (For now the main branch doesn't really play any particular role; it will only assume this once we've declared it to contain release 2.0, after which we won't touch it again until the next bump).

PuneetMatharu:
In the change-log text file, when you tell the reader that the third-party libraries (CGAL etc.) aren't supported on macOS, I think it might be worth quickly adding that the user has to supply the `--enable-suppress-cgal-build` flag in their configuration options to ensure this doesn't break their build. Thoughts?

MatthiasHeilManchester:
Ah, yes, I didn't know this. Can you add this in please? (Easier if you do it, because you know what exactly needs to be done; I may do something silly and check it because I have a proper computer). I think you can just edit the file from the webpage.

PuneetMatharu:
Done. How does that look?

Issue #41, opened 2021-09-07
Doxygen tweaks


MatthiasHeilManchester:
- [ ] Get rid of "My Project Documentation" on homepage...once laptop is updated to latest version; this must be created by a more recent version of doxygen. Happy for anybody else to fix this too.
- [ ] Get rid of all the \short things (Killing "\short" should do the trick) and set suitable flag in Doxyfile (MULTI_LINE_CPP_IS_BRIEF)
- [ ] Update instructions for how to write documentation accordingly.

puneetmatharu:
Task 1: "Get rid of 'My Project Documentation'..." ticked off the list (https://github.com/oomph-lib/oomph-lib/pull/42)!

PR #42, opened 2021-09-08
Fix auto-generated title in Doxygen documentation.

puneetmatharu:
Doxygen automatically generates the title "My Project Documentation" if you don't pass the `notitle` argument to the `\mainpage` command. I've now added the argument to fix the documentation. Details [here](https://www.doxygen.nl/manual/commands.html#cmdmainpage).

Addresses one of the tasks in issue https://github.com/oomph-lib/oomph-lib/issues/41.

### DIFF
--- a/doc/doc.txt
+++ b/doc/doc.txt
@@ -1,4 +1,4 @@
-\mainpage
+\mainpage notitle
 \htmlonly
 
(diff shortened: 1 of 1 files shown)


puneetmatharu:
Going to (again) do the naughty act of approving my own commit. Do as I say, and not as I do, eh...

PR #48, opened 2021-09-13
Added functionality implemented for a bubble in a Hele-Shaw channel by Jack Keeler

PaulAlexander23:
See the commit messages for further details.
There is a significant amount of code duplication however that is perhaps something to work on when someone has a little free time. I don't feel confident tackling it myself as this is primarily Jack's changes, where I have updated some of the comments.
Please could you check CRComplexMatrix::multiply in complex_matrices.cc. There are multiple approaches implemented where a single option is hard coded in. Perhaps the function could be pruned significantly.
Passes all the tests as far as I am aware.

PR #50, opened 2021-09-19
Patch issue with 'delete'-d solve() function of ComplexDampedJacobi and parallel_self_test.py

puneetmatharu:
## Description

Accidentally `delete`-d the broken virtual `solve()` function in `ComplexDampedJacobi` (that obviously shouldn't have been made to use `delete`). Credit to @PaulAlexander23 for spotting this mistake (thanks!) (see https://github.com/oomph-lib/oomph-lib/pull/48#discussion_r710429110). The self-tests didn't _appear_ to fail so I thought everything was fine. It turns out that some self-tests failed but the self-test workflow didn't recognize the error. It could be that the `parallel_self_test.py` script doesn't provide an exit code when some self-tests fail to build. Need to sort this out; will add it onto my to-do list.

**Example self-test workflow showing false-positive:** [latest macOS self-test workflow](https://github.com/oomph-lib/oomph-lib/runs/3600730006?check_suite_focus=true); see the number of `BUILD FAIL`s in the `Validate` step.

## To-do list

- [x] Fix issue with parallel self-test not recognising failing self-tests.
- [x] Wait for the self-tests to pass.
  - [macOS self-tests (passed)](https://github.com/oomph-lib/oomph-lib/runs/3600730006?check_suite_focus=true)
  - [Ubuntu self-tests (passed)](https://github.com/oomph-lib/oomph-lib/runs/3600730019?check_suite_focus=true)

### DIFF
--- a/bin/parallel_self_test.py
+++ b/bin/parallel_self_test.py
@@ -35,19 +35,25 @@ from __future__ import unicode_literals
 # x.encode()) / decoded (with str(y)) in python3. e.g. see
 # variable_from_makefile.
 
-
-import subprocess as subp
-from multiprocessing import Pool
-import sys
 import argparse
+import multiprocessing
 import os
 import os.path
-import multiprocessing
-import itertools as it
 import pprint
+import subprocess as subp
+import sys
 
+from enum import IntEnum
 from functools import partial as pt
+from multiprocessing import Pool
 from os.path import join as pjoin
+
+
+class ExitCode(IntEnum):
+    """Define our own custom enumeration to represent certain exit codes."""
+    SUCCESS = 0
+    BUILD_FAILURE = 1
+    TEST_FAILURE = 2
 
 
 class Colours:
@@ -197,7 +203,7 @@ def find_validate_dirs(base_dirs):
     return all_validation_dirs
 
 
-def dispatch_dir(dirname, features, **kwargs):
+def dispatch_dir(dirname, features, **kwargs) -> ExitCode:
     """Check for missing features and print the appropriate message if
     needed. Otherwise run the check function."""
 
@@ -209,8 +215,7 @@ def dispatch_dir(dirname, features, **kwargs):
             if feature['check_driver_function'](dirname):
                 missing_feature_message(dirname, feature['feature_name'])
                 return
-
-    make_check_in_dir(dirname, **kwargs)
+    return make_check_in_dir(dirname, **kwargs)
 
 
 # Functions for checking if a test needs a certain feature
@@ -230,7 +235,7 @@ def check_if_hlib_driver(d):
 
 # The function doing the bulk of the actual work (called many times in
 # parallel by main).
-def make_check_in_dir(directory, just_build=False):
+def make_check_in_dir(directory, just_build=False) -> ExitCode:
     """
     Rebuild binaries in the directory using make if needed then run the
     tests.
@@ -263,10 +268,15 @@ def make_check_in_dir(directory, just_build=False):
                                  stdout=tracefile,
                                  stderr=subp.STDOUT)
 
-        # If it failed then return a build failure immediately
-        if build_return != 0:
-            build_fail_message(directory)
-            return
+    # If the build failed then return with a build failure immediately
+    if build_return != ExitCode.SUCCESS:
+        build_fail_message(directory)
+        with open(tracefile_path, 'r') as tracefile:
+            print(tracefile.read())
+        return ExitCode.BUILD_FAILURE
+
+    # Re-open the tracefile and continue appending to it
+    with open(tracefile_path, 'a') as tracefile:
 
         if not just_build:
             tracefile.write("\nRunning self test properly:\n")
@@ -283,7 +293,6 @@ def make_check_in_dir(directory, just_build=False):
             tracefile.write(
                 "\nNot running self test because you set the 'just_build' option\n")
             tracefile.flush()
-
             test_result = 0
 
     final_tracefile_path = pjoin(directory, "Validation", "make_check_output")
@@ -301,12 +310,13 @@ def make_check_in_dir(directory, just_build=False):
 
     if test_result == 0:
         check_success_message(directory)
-        return
+        exit_code = ExitCode.SUCCESS
     else:
         with open(tracefile_name, 'r') as tracefile:
             print(tracefile.read())
         check_fail_message(directory)
-        return
+        exit_code = ExitCode.TEST_FAILURE
+    return exit_code
 
 
 # Function dealing with parsing of arguments, getting everything set up and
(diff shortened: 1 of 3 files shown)


PaulAlexander23:
Just some more info.
Initially found the issue on commit `84af2fc201 Tweak include paths for a few src/generic/ headers.` but probably was introduced earlier. Running `./bin/parallel_self_test.py` outputs 16 `BUILD_FAIL`. (We should probably log this somewhere rather than outputting to os.devnull).
Running `make check -k` in one of the failing directories outputs,
```
/home/paul/Repositories/oomph-lib/build/include/generic/linear_solver.h:152:18: note: overridden function is ‘virtual void oomph::LinearSolver::solve(oomph::Problem* const&, oomph::DoubleVector&)’
  152 |     virtual void solve(Problem* const& problem_pt, DoubleVector& result) = 0;
      |                  ^~~~~
make[1]: *** [Makefile:483: structured_cubic_point_source.o] Error 1
make[1]: Target 'structured_cubic_point_source' not remade because of errors.
make[1]: Leaving directory '/home/paul/Repositories/oomph-lib/demo_drivers/multigrid/helmholtz_multigrid'
make: *** [Makefile:688: check-am] Error 2
make: Target 'check' not remade because of errors.
```
But `./validate.sh` passes all the tests (the output executable files were built from a previous run on a working version)
```
/home/paul/Repositories/oomph-lib/build/include/pml_helmholtz/complex_smoother.h:486:10: error: deleted function ‘void oomph::ComplexDampedJacobi<MATRIX>::solve(oomph::Problem* const&, oomph::DoubleVector&) [with MATRIX = oomph::CRDoubleMatrix]’ overriding non-deleted function
  486 |     void solve(Problem* const& problem_pt, DoubleVector& result) = delete;
      |          ^~~~~
In file included from /home/paul/Repositories/oomph-lib/build/include/generic/assembly_handler.h:40,
                 from /home/paul/Repositories/oomph-lib/build/include/generic.h:9,
                 from structured_cubic_point_source.cc:33:
/home/paul/Repositories/oomph-lib/build/include/generic/linear_solver.h:152:18: note: overridden function is ‘virtual void oomph::LinearSolver::solve(oomph::Problem* const&, oomph::DoubleVector&)’
  152 |     virtual void solve(Problem* const& problem_pt, DoubleVector& result) = 0;
      |                  ^~~~~
make: *** [Makefile:483: structured_cubic_point_source.o] Error 1
Running multigrid validation with 2D Helmholtz problem
done
Running multigrid validation with 3D Helmholtz problem
done
 
======================================================================
 
All tests in
 
    /home/paul/Repositories/oomph-lib/demo_drivers/multigrid/helmholtz_multigrid    
 
passed successfully.
 
======================================================================
```
Let me know if you can't reproduce the build fails on that git commit and we can debug further.
I'm running `make check -k` globally now to see if that fails or produces a different output to `./bin/parallel_self_test.py`.

PaulAlexander23:
`make clean` before `make check -k` would ensure that the build errors are caught but would probably lead to greater testing times. Is this a faulty dependency thing? If the dependent files change, as in this case, surely we should rebuild the test program? (and if they haven't then we don't need to rebuild, to be efficient). Hopefully this will be fixed anyway with the move to cmake.

puneetmatharu:
Brilliant, thanks for the help @PaulAlexander23!  

> Initially found the issue on commit 84af2fc201 Tweak include paths for a few src/generic/ headers.

I'm pretty certain that this wasn't the commit that caused an issue; it very likely arose through an earlier commit when I added uses of `delete`. (...Oops.)

> Running ./bin/parallel_self_test.py outputs 16 BUILD_FAIL. (We should probably log this somewhere rather than outputting to `os.devnull`).

Absolutely, I suspected something like this would be the issue - the correct exit code isn't getting propagated to the workflow. Sorting this out is on my to-do list, but I'm not sure if I can get to it today. Maybe tomorrow!

> Is this a faulty dependency thing? If the dependent files change, as in this case, surely we should rebuild the test program?

I think the issue is with the `validate.sh` script. As you can see, when you run `./validate.sh`, it tries to re-`make` the executable, but fails due to an error with the build. As a result, an executable doesn't get generated and overwrite the previously existing one, and the old one gets run instead. I can think of two ways to address this:
1. The executable gets `rm -f`-ed before being re-built. With this approach, there will be no executable to run. Because `set -e` isn't enabled, I imagine the bash script will not fail when this happens and will just continue running. However, an `[OK]` won't get added to the `validation.log`. (So there will be an incorrect number of `[OK]`s in the end, which will flag the overall test up as a failure.)
2. Build an executable using `make $code || exit 1`; this will make the script exit early but it doesn't tell you whether other programs running in the same `validate.sh` script will also fail.

Because of the (undesirable) short-circuit logic of the second approach, I'd prefer to go with the former one (so that's what I'll add!).

PaulAlexander23:
Hmm. I don't like (1) personally as everything will be rebuilt rather than just the ones that need it (because the source code or their dependents have changed). It will work for sure, but if you do implement it could you time before and after the changes as speedy testing times is important!
I agree with your opinion of the second approach, although if we are going for passing all tests then one failing is sufficient to go back to debugging, but it will be less informative.
Does that make sense? I feel like I could be clearer.

PaulAlexander23:
Just to follow up on
> I'm running `make check -k` globally now to see if that fails or produces a different output to `./bin/parallel_self_test.py`.
All compiled tests passed successfully. Technically true with the inclusion of the 'compiled' part but doesn't actually test the source code!

MatthiasHeilManchester:
No No No! This is all wrong (and only happens in one demo_driver directory that somebody we know wrote, and somebody else we also know didn't check carefully). validate.sh is NOT supposed to build anything; it's run after the executables have been built (by make check). Happy to have a chat to explain; also I wouldn't mind a clarification of all this delete stuff which seems to have created a lot of problems (and I don't fully understand how it's supposed to work, etc.).

puneetmatharu:
Discussions were had, and wrists were slapped! 

The rule is "_no `validate.sh` script will, itself, build an executable_" because the `validate.sh` script should **_only_** be run (indirectly) via the `make check` command. The `make check` command will ensure previously-built executables are wiped before rebuilding, avoiding the scenario you described, @PaulAlexander23. 

The relevant changes have been made and committed. I suggest we wait until the self-tests pass before merging the changes in.

puneetmatharu:
Tests **before** fixing the `delete` specifier issue:
  - [macOS self-tests (fail)](https://github.com/oomph-lib/oomph-lib/runs/3600730006?check_suite_focus=true)
  - [Ubuntu self-tests (fail)](https://github.com/oomph-lib/oomph-lib/runs/3600730019?check_suite_focus=true) 

Tests **after** fixing the `delete` specifier issue:
  - [macOS self-tests (pass)](https://github.com/puneetmatharu/oomph-lib/runs/3666050847?check_suite_focus=true)
  - [Ubuntu self-tests (pass)](https://github.com/puneetmatharu/oomph-lib/runs/3666050841?check_suite_focus=true)

So it looks like the changes are all good to go!

puneetmatharu:
Tests **before** fixing the `delete` specifier issue:
  - [macOS self-tests (pass)](https://github.com/puneetmatharu/oomph-lib/runs/3666050847?check_suite_focus=true)
  - [Ubuntu self-tests (pass)](https://github.com/puneetmatharu/oomph-lib/runs/3666050841?check_suite_focus=true)

Tests **after** fixing the `delete` specifier issue:
  - [macOS self-tests (fail)](https://github.com/oomph-lib/oomph-lib/runs/3600730006?check_suite_focus=true)
  - [Ubuntu self-tests (fail)](https://github.com/oomph-lib/oomph-lib/runs/3600730019?check_suite_focus=true) 

Therefore the `delete` specifier issue is patched and the `parallel_self_test.py` script now propagates the correct exit code. Ready to be merged, boss!

PR #52, opened 2021-09-24
Move to C++11: Add support for construction of Vectors using initialiser_lists

puneetmatharu:
## Description

Adds some syntactic sugar for the construction of `Vector` objects. Makes use of `initialiser_list`s (not to be confused with constructor initialiser list), which are a C++11 construct.

### Original C++ code

Stolen from PR https://github.com/oomph-lib/oomph-lib/pull/48.

```cpp
Vector<std::complex<double>> values(4);
values[0] = std::complex<double>(1, 1);
values[1] = std::complex<double>(2, 1);
values[2] = std::complex<double>(3, 1);
values[3] = std::complex<double>(4, 1);

Vector<int> row_index(4);
row_index[0] = 0;
row_index[1] = 2;
row_index[2] = 1;
row_index[3] = 3;

Vector<int> column_start(3);
column_start[0] = 0;
column_start[1] = 2;
column_start[2] = 4;

Vector<std::complex<double>> values_square(7);
values_square[0] = std::complex<double>(1, 1);
values_square[1] = std::complex<double>(2, -1);
values_square[2] = std::complex<double>(3, 0.1);
values_square[3] = std::complex<double>(4, 3);
values_square[4] = std::complex<double>(5, 1);
values_square[5] = std::complex<double>(6, -2);
values_square[6] = std::complex<double>(0.5, -1);

Vector<int> row_index_square(7);
row_index_square[0] = 0;
row_index_square[1] = 2;
row_index_square[2] = 0;
row_index_square[3] = 1;
row_index_square[4] = 1;
row_index_square[5] = 2;
row_index_square[6] = 3;

Vector<int> column_start_square(5);
column_start_square[0] = 0;
column_start_square[1] = 2;
column_start_square[2] = 4;
column_start_square[3] = 6;
column_start_square[4] = 7;

Vector<std::complex<double>> x(4);
x[0] = std::complex<double>(2, 2);
x[1] = std::complex<double>(-2, 3);
x[2] = std::complex<double>(1, -2);
x[3] = std::complex<double>(-2, 1);

Vector<std::complex<double>> rhs(4);
rhs[0] = std::complex<double>(1, -0.5);
rhs[1] = std::complex<double>(0.5, 1.2);
rhs[2] = std::complex<double>(2, 0.3);
rhs[3] = std::complex<double>(-3, -0.4);
```

### Code using new constructor

```cpp
Vector<int> row_index{0, 2, 1, 3};
Vector<int> column_start{0, 2, 4};
Vector<std::complex<double>> values{{1, 1}, {2, 1}, {3, 1}, {4, 1}};

Vector<int> row_index_square{0, 2, 0, 1, 1, 2, 3};
Vector<int> column_start_square{0, 2, 4, 6, 7};
Vector<std::complex<double>> values_square{{1, 1}, {2, -1}, {3, 0.1}, {4, 3},
                                      {5, 1}, {6, -2}, {0.5, -1}};
                                      
Vector<std::complex<double>> x{{2, 2}, {-2, 3}, {1, -2}, {-2, 1}};
Vector<std::complex<double>> rhs{{1, -0.5}, {0.5, 1.2}, {2, 0.3}, {-3, -0.4}};
```

## Notes

- `initialiser_list` constructors are already available via direct access to other STL containers, e.g.

  ```cpp
  std::map<int, std::complex<double>>{
      {-3, {59.0, -1.0}},
      {20, {12.0, 5.0}},
      {5, {100.0, -9.0}},
      {299, {-1.0, 73.0}},
  };
  ```
  In this case, I just need to expose the underlying constructor in `std::vector` to the `Vector` wrapper class.
- You could also use this method to construct small dense matrices (i.e. as an `initialiser_list` of `initialiser_list`s), but I'm too lazy to do that myself.

## Self-tests

Still pass like normal:
- [macOS self-tests](https://github.com/puneetmatharu/oomph-lib/actions/workflows/self-tests-macos.yaml?query=branch%3Afeature-add-cxx11-vector-initialiser-constructor)**
- [Ubuntu self-tests](https://github.com/puneetmatharu/oomph-lib/actions/workflows/self-tests-ubuntu.yaml?query=branch%3Afeature-add-cxx11-vector-initialiser-constructor)**

_**click the badge to check that the most recent test has finished and passed._

### DIFF
--- a/src/generic/Vector.h
+++ b/src/generic/Vector.h
@@ -150,6 +150,13 @@ namespace oomph
     {
     }
 
+    /// \short A constructor that creates a vector with entries set by the
+    /// values in the input initialiser_list
+    /// Example:
+    ///           Vector<int> arr{0, 20, 100, 150);
+    ///           Vector<int> arr = {0, 20, 100, 150);
+    Vector(std::initializer_list<_Tp> init) : std::vector<_Tp>(init) {}
+
     /// Copy constructor
     Vector(const Vector<_Tp>& __x) : std::vector<_Tp>(__x) {}
 
@@ -231,6 +238,13 @@ namespace oomph
     Vector(size_type __n, const bool& __value) : std::vector<bool>(__n, __value)
     {
     }
+
+    /// \short A constructor that creates a vector with entries set by the
+    /// values in the input initialiser_list.
+    /// Example:
+    ///           Vector<int> arr{true, false, false, true);
+    ///           Vector<int> arr = {true, false, false, true);
+    Vector(std::initializer_list<bool> init) : std::vector<bool>(init) {}
 
     /// Copy constructor
     Vector(const Vector<bool>& __x) : std::vector<bool>(__x) {}


src/generic/Vector.h
```diff
@@ -150,6 +150,17 @@ namespace oomph
     {
     }
 
+    /// \short A constructor that creates a vector of variable size (using an
+    /// initialiser_list) with entries set to the values in the initialiser_list
```
jondea, line 154: I would say that `of variable size (using an initialiser_list)` is redundant and potentially confusing.

Also I seen you've added the ability to set the allocator, but a comment above says `The generalisation to general allocators is NOT handled here`. Might be worth checking with @MatthiasHeilManchester or @alhazel about whether this should still be the case.

jondea:
Oh I wish this had been done many many years ago ;) Thank you!

src/generic/Vector.h
puneetmatharu: Good point, I've stripped it from the comment. Thanks!

I hadn't noticed that - that's interesting one. I _did_ make an effort to avoid adding `std::allocator<T>` arguments everywhere else (as done with `std::vector`), so it will be the only constructor that takes an `std::allocator<T>`. Also, they may not want to generalise it to all possible STL constructors (e.g. move constructors), but I think this one is far too helpful to pass up!

puneetmatharu:
> Oh I wish this had been done many many years ago

So do I, @jondea. So do I...

src/generic/Vector.h
MatthiasHeilManchester: ??? Somebody will have to explain this to me!

puneetmatharu: **Allocators**

The `std::vector` class typically takes two values: (1) the element type, and (2) the allocator type (which has a default; `std::allocator<T>`). The `Vector` wrapper around `std::vector` drops the second template argument and only specifies the element type. Any calls to the underlying `std::vector` class therefore only need to specify the element type and the defaults specified by `std::vector` handle the allocator selection. I _think_ this is what the comment about "not generalising to general allocators" refers to.

After writing this out, I realise that I can actually omit the `std::allocator<_Tp>` argument, making the code even simpler, so I've changed that, and recommend waiting for the self-tests to pass.

**Initialiser lists**

An `initializer_list<T>` is a container introduced in C++11 that can be created using curly braces `{}` and can be passed any number of entries (of the same type!) and it will create a container of the appropriate size. Many (or possibly all; I'm not sure) STL containers have a constructor that allows a user to initialise using an `initializer_list<T>`. It also lets users use a new type of initialisation called "list initialisation" which does not allow narrowing conversions (which can be quite nice), e.g.

```cpp
// Okay
int a{1};

// Error!
// temp.cc:3:9: error: type 'double' cannot be narrowed to 'int' in initializer list [-Wc++11-narrowing]
//  int a{1.0};
int a{1.0};
```
More details
- [modern-cpp-features](https://github.com/AnthonyCalandra/modern-cpp-features#initializer-lists) 
- [list initialisation](https://en.cppreference.com/w/cpp/language/list_initialization).

MatthiasHeilManchester:
> Oh I wish this had been done many many years ago ;) Thank you!

Using heart emoticons in hardcore coding discussions? Tssk, tssk, tssk. But yeah, all good stuff, and the executive summary is that you (Puneet) have added the functionality to oomph-lib's Vector to make these initialisations possible, right?

src/generic/Vector.h
MatthiasHeilManchester: > Good point, I've stripped it from the comment. Thanks!

Did you? Or rather did you do it and then push your change? Can't see it!

puneetmatharu: ...not sure what you're talking about, it's all definitely changed 👀

puneetmatharu:
> Using heart emoticons in hardcore coding discussions? Tssk, tssk, tssk. 

How else am I meant to positively-reinforce people agreeing with me? 

> ...and the executive summary is that you (Puneet) have added the functionality to oomph-lib's Vector to make these initialisations possible, right?

Correctamundo. So you know where to go if there's any `git blame`-ing to be done!

P.S. I'm running the self-tests one last time after changing the missed comment, so I'd recommend waiting until they pass before merging the request in! (I've updated the main PR message to display the corresponding self-test status badges.)

PR #53, opened 2021-09-30
Patch hypre; remove some check for headers; clarify doc

MatthiasHeilManchester:
- Omitted some checks for headers in configure.ac since the test fails even though the headers are there. The problem appears to be that the headers can't be compiled independently -- don't quite understand the problem, and gave up trying to understand the logic...
- Patched hypre in the build script to make sure it works with the latest version of openmpi (original code uses deprecated mpi interfaces). Note that this isn't actually checked in the current github self-tests but the full mpi-based self-tests pass on my office machine which I've updated to ubuntu 20.04. Am in the process of writing an "Action" that does the self-tests for mpi too.
-  Clarified naming rules for variables and access functions related to the number of entries in a container (in coding conventions)

### DIFF
--- a/doc/coding_conventions/coding_conventions.txt
+++ b/doc/coding_conventions/coding_conventions.txt
@@ -116,7 +116,11 @@ too much.
    /// Total number of elements
    unsigned nelement();
    \endcode
- \b Notes: (i) No underscore between the "n" and the container's name.
+ \b Notes: (i) No underscore between the "n" and the container's name;
+ You can then use the underscore for the name of the local variable
+ that stores the value, e.g.
+ \c
+ unsigned \c n_element \c = \c my_mesh_pt->nelement();
 (ii) No trailing \c _pt in the function that returns the number
 of objects in the container.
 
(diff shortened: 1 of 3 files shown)


doc/coding_conventions/coding_conventions.txt
```diff
@@ -116,7 +116,11 @@ too much.
    /// Total number of elements
    unsigned nelement();
    \endcode
- \b Notes: (i) No underscore between the "n" and the container's name.
+ \b Notes: (i) No underscore between the "n" and the container's name;
+ You can then use the underscore for the name of the local variable
+ that stores the value, e.g.
+ \c
+ unsigned \c n_element \c = \c my_mesh_pt->nelement();
```
puneetmatharu, line 123: Lowercase "y" in "you" on the second line?

puneetmatharu:
> ...Am in the process of writing an "Action" that does the self-tests for mpi too.

I've got a feature branch (that's in the works) containing updates to the workflows to run MPI self-tests too. However, there are currently a bunch of self-tests that mysteriously die on macOS, but don't on Ubuntu... It's probably worth us having a chat about this; I'll send you an email

doc/coding_conventions/coding_conventions.txt
MatthiasHeilManchester: Yes sir! :) And thanks for adding the labels; I agree that they are really useful but I keep forgetting.

PR #58, opened 2021-10-02
Add MPI testing for Ubuntu to self-test workflows

puneetmatharu:
## Description

Adds support for testing MPI-enabled demo drivers on Ubuntu via the GitHub Action self-tests.

## Notes

- Added a very minor tweak to the macOS self-test workflow to tidy it up a little.
- Updated the names of the uploaded `validation.log` artefacts. For the Ubuntu self-test workflow, this will ensure that the log files don't overwrite each other when uploaded.
- The non-MPI-enabled job and the MPI-enabled job are combined into a single workflow using a `matrix`.
- The PR does **not** add an extra status badge to the `README.md` to display the status of MPI-enabled tests on Ubuntu because status badges are provided "per workflow/branch" rather than "per workflow job".
- Does not provide support for testing MPI-enabled demo drivers on macOS due to mysterious MPI-related seg. faults in the following demo drivers:
  1. `demo_drivers/mpi/distribution/adaptive_driven_cavity`
  1. `demo_drivers/mpi/distribution/bifurcation_tracking`
  1. `demo_drivers/mpi/distribution/hp_adaptive_driven_cavity`
  1. `demo_drivers/mpi/distribution/three_d_entry_flow`
  1. `demo_drivers/mpi/multi_domain/boussinesq_convection`
  1. `demo_drivers/mpi/multi_domain/turek_flag`

Four of these six failing drivers (ii, iv, v, vi) appear to die somewhere in `superlu_dist_distributed_matrix`. The other two (i and iii) die somewhere in the adaptive Newton solve.

You can find the [failing macOS self-tests output here](https://github.com/puneetmatharu/oomph-lib/actions/runs/1298069117). I've also attached the corresponding log-file here: [`validation-macOS-MPI_true.log`](https://github.com/oomph-lib/oomph-lib/files/7271668/validation-macOS-MPI_true.log).

## Self-tests

Below are the status badges for Ubuntu with and without MPI, and macOS with no MPI.

- [Ubuntu self-tests](https://github.com/puneetmatharu/oomph-lib/actions/workflows/self-tests-ubuntu.yaml?query=branch%3Afeature-add-mpi-testing-to-self-test-workflows)*
- [macOS self-tests](https://github.com/puneetmatharu/oomph-lib/actions/workflows/self-tests-macos.yaml?query=branch%3Afeature-add-mpi-testing-to-self-test-workflows)*

*_Once these status badges become green, both tests have passed. You can also just click on them to see the test progress._

### DIFF
--- a/.github/workflows/self-tests-macos.yaml
+++ b/.github/workflows/self-tests-macos.yaml
@@ -10,9 +10,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Install requirements
-        run: |
-          brew install make automake libtool autoconf doxygen
-          brew reinstall gfortran
+        run: brew install make automake libtool autoconf doxygen && brew reinstall gfortran
 
       - name: Extra config
         run:
@@ -32,7 +30,7 @@ jobs:
       - name: Upload validation log file
         uses: actions/upload-artifact@v2
         with:
-          name: macos-latest validation log
+          name: validation-${{ runner.os }}.log
           path: ./validation.log
 
       - name: Propagate self-test status
--- a/.github/workflows/self-tests-ubuntu.yaml
+++ b/.github/workflows/self-tests-ubuntu.yaml
@@ -4,15 +4,33 @@ on: push
 
 jobs:
   build:
+    strategy:
+      fail-fast: false
+      matrix:
+        # Each map in the 'mpi_options' array contains a flag indicating whether
+        # to enable MPI and the name of the configuration file required for the
+        # MPI case, respectively
+        mpi_options: [
+          {enable: false, config_file: default},
+          {enable: true,  config_file: mpi_default_ci}
+        ]
+
     runs-on: ubuntu-latest
     steps:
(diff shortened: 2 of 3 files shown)


config/configure_options/mpi_default
```diff
@@ -0,0 +1,103 @@
+#=====================================================================
```
MatthiasHeilManchester, line 1: This is fine, apart from the name which suggests that it's the default that everybody should use to compile with mpi support. Not true; the oversubscribe flag is something specific that's only there for the github self tests so shouldn't this file be called mpi_for_github_selftests, say?

puneetmatharu: I initially thought that the `--oversubscribe` flag wouldn't change the performance of programs that are not oversubscribed, but [that's not true](https://www.open-mpi.org/faq/?category=running#oversubscribing), so you make a good point. I've changed the name to `mpi_default_ci`.

PR #63, opened 2021-10-19
Update urls for new location of external distribution files

MatthiasHeilManchester:
Specifically:

- third party distributions (hypre, trilinos and mumps) tar files
now live at https://personal.maths.manchester.ac.uk/oomphlib/oomph-lib_external_distfiles/

- Have also added tar file with bleeding edge doxygen distribution to that site, so we don't rely on it being kept available on the doxygen webpage.

### DIFF
--- a/bin/change_urls.bash
+++ b/bin/change_urls.bash
@@ -1,0 +1,40 @@
+#! /bin/bash
+
+#-----------------------------------------------------------
+# Search through all files and replace one url with another
+# Ft. Puneet magic to escape awkward characters...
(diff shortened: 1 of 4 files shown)


jondea (changes requested):
The URL changes look fine, but I would suggest that we probably don't need the script now that it's done it's job. If it's just a global find and replace, then it seems more like a personal script, rather than in oomph-lib.

jondea:
I also can't see any reference to the Doxygen tar file in the diff, is this intentional?

PR #70, opened 2021-11-15
Added formatting helper function to TimingHelpers (oomph_utilities.cc)

wobblygomboc:


src/generic/oomph_utilities.cc
```diff
@@ -1306,6 +1306,41 @@ namespace oomph
         return double(t) / double(CLOCKS_PER_SEC);
       }
     }
+
+    /// Returns a nicely formatted string from an input time in seconds;
+    /// the format depends on the size of time, e.g.:
+    ///  3710 will be printed as 1h:01:50
+    ///   700 will be printed as 11m:40
+    ///    59 will be printed as 59s
+    std::string format_time(const double& time)
```
puneetmatharu, line 1315: Might be more meaningful to name this something like `get_time_as_string(...)`

src/generic/oomph_utilities.cc
```diff
@@ -1306,6 +1306,41 @@ namespace oomph
...
+    {
+      std::ostringstream ss;
+  
+      unsigned total_time = unsigned(time);
+      unsigned time_hrs = unsigned(total_time) / 3600;
+      unsigned time_min = (unsigned(total_time) % 3600) / 60;
+      unsigned time_sec = (unsigned(total_time) % 3600) % 60;
```
puneetmatharu, line 1322: Are you sure you need the explicit `unsigned()` casts on the RHS? I don't think you do...

src/generic/oomph_utilities.h
```diff
@@ -952,6 +952,13 @@ namespace oomph
     /// returns the time in seconds after some point in past
     double timer();
 
+    /// Returns a nicely formatted string from an input time in seconds;
+    /// the format depends on the size of time, e.g.:
+    ///  3710 will be printed as 1h:01:50
+    ///   700 will be printed as 11m:40
+    ///    59 will be printed as 59s
+    std::string format_time(const double& time);
```
puneetmatharu, line 960: Will need to be updated if the function name changes

src/generic/oomph_utilities.cc
puneetmatharu: Also, I think the input should be something like a `long long unsigned` or `uint64_t` (after adding `#include <cstdint>`) given that you don't handle quantities smaller than a single second (i.e. `1`)

src/generic/oomph_utilities.cc
```diff
@@ -1306,6 +1306,41 @@ namespace oomph
...
+      else if (time_min > 0)
+      {
+        ss << time_min << "m:";
+        ss << std::setw(2) << std::setfill('0');
+        ss << time_sec << "\n" << std::endl;
+      }
+      else
+      {
+        ss << time_sec << "s\n" << std::endl;
+      }
```
puneetmatharu, line 1340: Do you think it might look nicer to print something like `3712` as `1h 01m 52s`? E.g. using something like
```cpp

// Include the headers file
if (time_hrs > 0)
{
  ss << time_hrs << "h ";
}
if (time_min > 0)
{
  ss << time_min << "m ";
}
ss << time_sec << "s\n" << std::endl;
```

puneetmatharu (changes requested):
Quite a neat addition, thanks for adding it in! I've only got a few small remarks on it, let me know what you think

MatthiasHeilManchester:
> Also, that's a quite deceptive PR name/description given that 90% is dedicated to adding in a helper function into `TimingHelpers`. Sneaky! (Not getting past me that easily :) )

Nobody would ever dream of trying to get anything sneaky past you, @puneetmatharu . But yes, we should probably update the title of the pull request to mention the TimingHelpers (especially since it is useful functionality!). Agree with Puneet's other queries too.

src/generic/oomph_utilities.cc
wobblygomboc: agreed

src/generic/oomph_utilities.cc
wobblygomboc: good point, total_time is already unsigned. And I can remove that variable entirely once I change the function argument type to unsigned

src/generic/oomph_utilities.cc
wobblygomboc: I guess the format is a bit of personal preference. But good point on not terminating the string, the user shouldn't be forced into having a newline appended.

src/generic/oomph_utilities.cc
```diff
@@ -1306,6 +1306,50 @@ namespace oomph
...
+        ss << std::setw(2) << std::setfill('0');
+        ss << time_sec;
+      }
+      else
+      {
+        ss << time_sec << "s";
+      }
+
+      return ss.str();
+    }
```
puneetmatharu, line 1352: This can probably be simplified using the built-in `sprintf` function.

```cpp
// Type alias for unsigned long long (will locally override the "time_t" in <ctime>)
using time_t = unsigned long long;    

// -------------------------------------- TODO: Delete
#include <climits>
time_t time_in_sec = ULLONG_MAX;
// -------------------------------------- TODO: Delete

// Calculate quantities
time_t days = time_in_sec / (3600 * 24);
time_t hours = (time_in_sec % (3600 * 24)) / 3600;
time_t minutes = (time_in_sec % 3600) / 60;
time_t seconds = time_in_sec % 60;
    
// Comment out this #define to see the output of the other block
#define WHITESPACE_DELIMETER
#ifdef WHITESPACE_DELIMETER

// This version is clear due to the visual whitespace break between quantities.
// Example: ULLONG_MAX -> "213503982334601h 07h 00m 15s".
char time[100];
sprintf(time, "%lluh %02lluh %02llum %02llus", days, hours, minutes, seconds);
std::string time_as_string(time);

#else

// Slightly harder to read due to the lack of a visual delimeter in the second 
// set of quantities.
// Example: ULLONG_MAX -> "213503982334601h 07h:00m:15s".
char time[100];
sprintf(time, "%lluh %02lluh:%02llum:%02llus", days, hours, minutes, seconds);
std::string time_as_string(time);

#endif

// -------------------------------------- TODO: Delete
std::cout << time_as_string << std::endl;
// -------------------------------------- TODO: Delete
```

puneetmatharu: The above looks slightly messy because of the comparison but the finished product would look like the following

```cpp
std::string convert_secs_to_formatted_string(const unsigned long long& time_in_sec) {
    // Type alias for unsigned long long (will locally override the "time_t" in <ctime>)
    using time_t = unsigned long long;    
    
    // Calculate quantities
    time_t days = time_in_sec / (3600 * 24);
    time_t hours = (time_in_sec % (3600 * 24)) / 3600;
    time_t minutes = (time_in_sec % 3600) / 60;
    time_t seconds = time_in_sec % 60;
            
    // This version is clear due to the visual whitespace break between quantities.
    // Example: ULLONG_MAX -> "213503982334601h 07h 00m 15s".
    char time[100];
    sprintf(time, "%lluh %02lluh %02llum %02llus", days, hours, minutes, seconds);
    return std::string{time};
}
```

wobblygomboc: I have to disagree - with this implementation, convert_secs_to_formatted_string(1) would print:
0d 00h 00m 01s
which is unnecessarily hard to read, whereas my implementation would print:
1s
 - I think we should prioritise the elegance of the output over the elegance of the code in this instance since it's specifically a function which gives nicely formatted output!

puneetmatharu: Ahhh, good point, my mistake there!

puneetmatharu (approved):

PR #72, opened 2021-11-22
Check free surface docs

alhazel:
Update two comments in one header file so that the Doxygen documentation works again. 

These comments were hooks and their removal / modification broke the documentation.


alhazel:
Aargh. I see that the clang format has changed things and I'll have to update!

PR #76, opened 2021-12-20
Patch for broken Doxygen formulae incl. update of .clang-format

puneetmatharu:
## Issue

Apparently recent versions of Doxygen cannot handle latex equations that span multiple lines. So, something like this

```cpp
/// the following equation
/// \f[
/// y = x
/// \f]
/// \f$ y = x
/// \f$
/// bla bla
```

will generate the error

```bash
warning: End of comment block while inside formula.
```

and documentation of the relevant class/function/... will either be badly formatted or the class may not be documented at all.

## Solution

The solution required two changes: (1) fixing the C++ comment blocks, and (2) updating the `.clang-format` file. 

#### 1. Fixing the C++ comment blocks

To fix the comment blocks, entire formulae need to be placed on a single line, e.g.

```cpp
/// the following equation
/// \f[ y = x \f]
/// \f$ y = x \f$
/// bla bla
```
 
In future, latex equations should be formatted in that manner. The reformatting of existing code was done with the new Python script `fix_broken_doxygen_formulae.py`, which has been placed in the `bin/` directory. This can be re-run occasionally to fix any violations of the above rule that may creep in. Run the script from the root `oomph-lib` directory using the following command:

```bash
# To print the possible command-line options
python3 ./bin/fix_broken_doxygen_formulae.py -h

# Format all *.{h,cc} files in the src/ directory and edit files in-place
python3 ./bin/fix_broken_doxygen_formulae.py --dir ./src --in-place
```

#### 2. Updating the `.clang-format` file

To ensure `clang-format` doesn't reflow comments so that they adhere to the 80-column-limit rule, I've specified a `CommentPragma`:

```yaml
CommentPragmas:  '(\\f\[.*\\f\]|\\f\$.*\\f\$)'
```

which is a regex pattern that identifies code that should not be broken onto multiple lines. With the above specification, formulae of the form `\f[ ... \f]` or `\f$ ... \f$` will be left alone. The updated version of the .clang-format file is part of this pull request.

## Self-tests

[macOS self-tests](https://github.com/puneetmatharu/oomph-lib/actions/runs/1601837857)
[Ubuntu self-tests](https://github.com/puneetmatharu/oomph-lib/actions/runs/1601837856)


MatthiasHeilManchester:
Have updated the description of the pull request slightly. Still happy with it?

puneetmatharu:
Yep! Thanks for doing that. I've also added instructions on how to run the Python script

PR #77, opened 2021-12-21
Pull to main in anticipation of version 2.0 release?

MatthiasHeilManchester:
Is this it? Version 2.0? Can't see any outstanding issues that need urgent attention so we may as well pull it across into main now.


puneetmatharu:
Having previously looked at most of these new commits, I think these changes are good to go (but it's still worth getting the sign-off from Andrew and Jon). 

However, I would recommend that we fix the issue with the deployment of the docs before merging the PR in, to avoid having to deal with bugs on the `main` branch

alhazel:
Happy to sign off. I haven't been through all the changes, but all seems generally sensible and the tests pass. I see that my very minor tweaks have been incorporated.

PR #78, opened 2021-12-28
[PATCH] Ubuntu MPI self-tests don't actually install MUMPS/ScaLAPACK

puneetmatharu:
## Problem

@MatthiasHeilManchester spotted that MUMPS and ScaLAPACK weren't actually installed during the Ubuntu MPI self-tests. Example [here](https://github.com/oomph-lib/oomph-lib/runs/4598608734?check_suite_focus=true). Below is a snippet of the output during the "Build" step showing why the issue arises; the build can't find `mpi.h`.

```bash
161 mpi include directory (the directory that contains mpi.h) 
162 has been specified with --with-mpi-include-directory
163 therefore I can do mumps/scalapack build from source
164 Not found mpi.h for scalapack build
165 Did not find mpi.h in  for scalapack build
166 Found MUMPS_4.10.0.tar.gz
167 Found scalapack_installer.tgz
168 Not doing mumps/scalapack build because location of mpi.h file not specified
```

## Solution 

I've added a step to the workflow to find the MPI include directory (specifically the one containing the `mpi.h` file) using `mpicc --showme:incdirs`. The GitHub runner has two include directories so we have to loop over them to find the directory we want. (Only one of these directories actually holds the `mpi.h` header.) The MPI header include path then gets passed to `oomph-lib` using the `--with-mpi-include-directory` flag, which is added to the MPI config file. The additional "Print updated config file (without comments)" step is there to help with debugging if anything goes wrong in the future.

## Self-test

- [Ubuntu self-tests](https://github.com/puneetmatharu/oomph-lib/actions/runs/1630347303)

### DIFF
--- a/.github/workflows/self-tests-ubuntu.yaml
+++ b/.github/workflows/self-tests-ubuntu.yaml
@@ -12,7 +12,7 @@ jobs:
         # MPI case, respectively
         mpi_options: [
           {enable: false, config_file: default},
-          {enable: true,  config_file: mpi_default_ci}
+          {enable: true, config_file: mpi_default_ci}
         ]
 
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


MatthiasHeilManchester:
Brilliant. Many thanks for sorting this out so quickly!

PR #80, opened 2022-01-05
Change to eigensolver interfaces (and functionality)

MatthiasHeilManchester:
@alhazel : this is one for you to check and sign off. Assuming you're happy with these new interfaces, you should presumably revisit the legacy versions of these functions at some point to either get rid of them and/or make sure that what they're computing is well-defined and clear to the user; see below for my reservations.

Self-tests pass on github and my laptop and the version is synced with 2.0.0.

UPATED ON JANUARY 15 TO DEAL WITH ALL ISSUES RAISED

## Summary of changes

  - Added postfix "_legacy" to existing problem-based eigensolver (and associated
    linear-algebra-type functions). Andrew needs to double check these
    at some point. The functions returned vectors of doubles which were
    in some (solver-specific, but not transparent) way related to the
    complex eigenvectors. I strongly suspect that the eigenvectors
    returned from anasazi may occasionally have been interpreted wrongly 
    since the index vector that clarifies what's what was never interrogated.
  
  - New eigensolvers (problem and linear-algebra based) now return
    - complex eigenvectors and
    - either:
      - complex eigenvalues (dangerous) or
      - complex eigenvalues encoded as fractions; this allows the
        NaN-free handling of infinite eigenvalues by the user.
      A convenience default version of the function that returns
      complex eigenvalues (which may be NaNs!) is provided; it either
      uses what is returned from the underlying linear algebra
      eigensolver (and is therefore likely to be safe, e.g. in Arnoldi methods)
      or brutally performs the division, NaN or not...

 - Eigenvectors are returned in the form of two DoubleVectors, one of which 
   contain the real part, the other the imaginary part of the (in general) complex eigenvectors.
  
  - Added error checks (monitoring a non-zero info flag) to lapack qz wrappers.
  
  - Added cc-ness check to eigenvalues in lapack qz wrappers (incl. some
    tolerance) to make sure the real vectors returned from this function
    are interpreted correctly and copied correctly into complex eigenvectors.
  
  - Replaced separately defined shift in Anasazi eigensolver by version
    already defined in base class.
  
  - Enclosed doc of memory statistics in Superlu in Doc_stats-based if statement.
 
  - Removed ARPACK as the default eigensolver but kept the code alive
     so Andrew can re-check things when updating "his" demo-driver codes
     and tutorials. New default is lapack's QZ and the tests all pass.
    
       NOTE: Andrew will have to sort out some of the self-tests because
             I don't understand the output from the legacy version of the
             eigensolvers well enough to update this. When validation data
             came from arpack I've now overwritten this with output from
             lapack QZ. In some codes (where all available eigensolvers
             were tested) this data is now output twice; should be cleaned
             but will require update of the validata and validate.sh.
    
       NOTE: Self test data from several anasazi based solves doesn't
             appear to be included in the validata (comenting out the
             relevant solves still makes the self-test pass). Odd.
    
   - Updated Makefile.am s which bypassed validation directories based on
     ARPACK's availability.
    
   - Andrew's Orr Sommerfeld code is now done with anasasi (computing all
      eigenvalues with QZ didn't work, presumably because of non-finite
      eigenthings... Makefile.am modified to check the availability of
      trilinos. This replaces the test for the availability of arpack.
    
   - Updated metis-based partitioning to handle the case when none of the
      elements in the Problem's mesh are connected (typically arises only
      in (dummy?) problems where all elements have only internal Data.
    
   - Added eigenvalues-only legacy eigensolver. (And that one has the
      same interface as a non-legacy version so we do need to retain the
      modifier.    

 - Andrew used of the Dof_distribution_pt (which describes the
      distribution of dofs over processors and therefore results in
      non-distributed objects if the Problem isn't distributed -- it
      usually isn't, and we still want to benefit from maximum parallel
      goodness!) in eigenproblems. This resulted in odd behaviour because
      the eigenvectors, say, weren't distributed, whereas all the other
      linear algebra-type things computed by the Problem (Jacobian,
      residuals, mass matrix, etc.) were. This was done because he
      couldn't get hold of the  linear algebra distribution used for the
      linear algebra. Turns out this was set up (with lots of code
      duplication whereever it was needed but was indeed not accessible
      to the outside world.
    
      Have now created a new function
    
        void Problem::create_new_linear_algebra_distribution(
                     LinearAlgebraDistribution*& dist_pt)
    
      that does the honours (and is accessible from the outside).
    
      With this the (new) eigensolvers (not the _legacy ones) now return
      distributed eigenvectors if the code is run on multiple processors.
    
  - introduced temporary boolean
    
         Anasazi::Use_temporary_code_for_andrew_legacy_version
    
      to keep code required for the _legacy eigensolvers alive.
      Don't understand why the legacy anasazi version doesn't work with the
      updated operators (implemented in oomph-lib to make anasazi work)
      and don't understand Andrew's driver code well
      enough to understand what he's doing when I replace the legacy
      version with the new one (which does work in problems that I do
      understand.
    
  - Changed the verbosity of anasazi eigensolver -- I felt it cluttered
      up the screen too much. Whoever wants the full information back
      should probably give the user access to the level of verbosity via
      suitable member functions.

- Commented out many member functions (and associated private data)
  in Anasazi eigensolver. They were neither used nor required from the
  Eigensolver base class (though they seemed useful!). I have no use case
  for them so can't check them but if they're required (in Anasazi's anticipated
  replacement for arpack) then they should be reimplemented in such a way  
  that they actually do what they say.


puneetmatharu:
Should the legacy method be avoided (i.e. it's deprecated) or should people still be allowed to use it? 

If it's the latter, we could bump up to C++14 and use a "deprecated" attribute (completely optional):

```cpp
// Original
void ARPACK::solve_eigenproblem_legacy(/* ... */);

// Updated
[[deprecated("Use the implementation of solve_eigenproblem() taking alpha and beta instead.")]]
void ARPACK::solve_eigenproblem(/* ... */)
```

We don't need to worry about the conflicting function names because the function signature is different ([example](cpp.sh/5ze7t)). With the added attribute, users of the legacy method get a compiler warning about the deprecated function of the form:

```bash
eigen_solver.h:21:5: warning: 'solve_eigenproblem' is deprecated: Use the implementation of solve_eigenproblem() taking alpha and beta instead. [-Wdeprecated-declarations]
    solve_eigenproblem();
    ^
eigen_solver.h:8:3: note: 'solve_eigenproblem' has been explicitly marked deprecated here
[[deprecated("Use the implementation of solve_eigenproblem() taking alpha and beta instead.")]]
```

MatthiasHeilManchester:
> Should the legacy method be avoided (i.e. it's deprecated) or should people still be allowed to use it?

For now this is a maths question, and one primarily for @alhazel -- not sure he agrees that it's legacy (or deprecated!). He may want to keep the interface alive (and then come up with some other name (and a proper description of what it does) for it).

bin/emacs_files_that_contain_string.bash
```diff
@@ -21,7 +21,7 @@ echo " "
 fi
 
 #Search string 
-search_string=RefineableTriangleMesh
+search_string=solve_eigenproblem
```
alhazel, line 24: As a note, these files do not necessarily need to be updated in the repository. There was a way to do this in svn that we never really used and I don't feel strongly enough about it to investigate whether a similar trick is possible in git.

src/generic/eigen_solver.cc
```diff
@@ -623,7 +641,7 @@ namespace oomph
     int info = 0;
 
     // Call FORTRAN LAPACK to get the required workspace
-    // Note the use of the padding flag
+    // Note the use of the padding leading dimension for the matrices
```
alhazel, line 644: Not clear where the padding leading dimension is used in the call. n is the original dimension isn't it?

src/generic/eigen_solver.cc
```diff
@@ -666,26 +691,37 @@ namespace oomph
...
       // Resize the eigenvector  storage
-      eigenvector[i].build(this->distribution_pt(), 0.0);
-      // Load up the eigenvector (assume that it's real)
+      eigenvector_aux[i].build(this->distribution_pt(), 0.0);
+
+      // Load up the mangeled storage for the eigenvector. What's called
+      // eigenvector here isn't necessarily an eigenvector but provides storage
+      // for the real and imaginary parts of the real or complex conjugate
+      // eigenvectors. They are translated into actual complex eigenvectors
+      // (at the cost of some additional storage and the gain of considerable
```
alhazel, line 720: gain of considerable what?

src/generic/eigen_solver.cc
```diff
@@ -700,19 +736,205 @@ namespace oomph
...
+    //        v(j) = VR(:,j)+i*VR(:,j+1) and v(j+1) = VR(:,j)-i*VR(:,j+1).
+    //        Each eigenvector is scaled so the largest component has
+    //        abs(real part)+abs(imag. part)=1.
+    unsigned n = problem_pt->ndof();
+    eigenvector.resize(n);
+    unsigned eval_count = 0;
+    while (eval_count < n)
+    {
+      // i-th eigenvalue is real:
+      if (alpha_eval[eval_count].imag() == 0.0)
```
alhazel, line 819: Should there be a threshold in here given that this is a floating point comparison?

bin/emacs_files_that_contain_string.bash
puneetmatharu: Good point

@MatthiasHeilManchester, see [here](https://stackoverflow.com/a/3320183) for info on how precisely to do this

**TL;DR:** Use 

```bash
git update-index --assume-unchanged bin/emacs_files_that_contain_string.bash
```

to tell Git to not track changes on that file as long as you want. When you want to make Git track it again, use

```bash
git update-index --no-assume-unchanged bin/emacs_files_that_contain_string.bash
```

You will encounter a Git error when you run `git pull` if you have modified that file. In this case, just undo the changes using 

```bash
git checkout bin/emacs_files_that_contain_string.bash
```

or just stash the changes, pull in the changes and unstash

src/generic/eigen_solver.cc
```diff
@@ -700,19 +736,205 @@ namespace oomph
...
+        if ((beta_eval[eval_count] != 0.0) &&
+            (beta_eval[eval_count + 1] != 0.0))
+        {
+          std::complex<double> lambda_this =
+            alpha_eval[eval_count] / beta_eval[eval_count];
+
+          std::complex<double> lambda_next =
+            alpha_eval[eval_count + 1] / beta_eval[eval_count + 1];
+
+          // Check cc-ness to within tolerance
```
alhazel, line 848: Modify comment to indicate that this is a check for failure of cc-ness?

src/generic/eigen_solver.cc
```diff
@@ -700,19 +736,205 @@ namespace oomph
...
+            error_stream
+              << "isn't the negative of its subsequent value       : "
+              << lambda_next.imag() << std::endl
+              << "Their sum " << (lambda_this.imag() + lambda_next.imag())
+              << " is greater than Tolerance_for_ccness_check = "
+              << Tolerance_for_ccness_check << std::endl;
+            throw OomphLibError(error_stream.str(),
+                                OOMPH_CURRENT_FUNCTION,
+                                OOMPH_EXCEPTION_LOCATION);
+          }
```
alhazel, line 865: throw separate exceptions or combine and throw one exception?

src/generic/eigen_solver.cc
```diff
@@ -700,19 +736,205 @@ namespace oomph
...
+              << "Their difference "
+              << (lambda_this.real() - lambda_next.real())
+              << " is greater than Tolerance_for_ccness_check = "
+              << Tolerance_for_ccness_check << std::endl;
+            throw OomphLibError(error_stream.str(),
+                                OOMPH_CURRENT_FUNCTION,
+                                OOMPH_EXCEPTION_LOCATION);
+          }
+        }
+        else
```
alhazel, line 884: Should this else block be removed?

src/generic/eigen_solver.cc
```diff
@@ -700,19 +736,205 @@ namespace oomph
...
+    //        their eigenvalues. If the j-th eigenvalue is real, then
+    //        v(j) = VR(:,j), the j-th column of VR. If the j-th and
+    //        (j+1)-th eigenvalues form a complex conjugate pair, then
+    //        v(j) = VR(:,j)+i*VR(:,j+1) and v(j+1) = VR(:,j)-i*VR(:,j+1).
+    //        Each eigenvector is scaled so the largest component has
+    //        abs(real part)+abs(imag. part)=1.
+    unsigned n = problem_pt->ndof();
+    eigenvector.resize(n);
+    unsigned eval_count = 0;
+    while (eval_count < n)
```
alhazel, line 816: This will always return n eigenvalues not neval. Is that what we want? I don't know that I have an opinion.

src/generic/eigen_solver.h
```diff
@@ -74,28 +74,92 @@ namespace oomph
...
+    /// Solve the real eigenproblem that is assembled by elements in
+    /// a mesh in a Problem object. Note that the assembled matrices include the
+    /// shift and are real. The eigenvalues and eigenvectors are,
+    /// in general, complex, and the eigenvalues can be NaNs or Infs.
+    /// This function is therefore merely provided as a convenience, to be
+    /// used if the user is confident that NaNs don't arise (e.g. in Arnoldi
+    /// based solvers where typically only a small number of (finite)
+    /// eigenvalues are computed), or if the users is happy to deal with NaNs in
+    /// the subsequent post-processing.
+    /// Function is virtual so it can be overloaded for arnoldi type solvers
```
alhazel, line 102: arnoldi should be Arnoldi

src/generic/eigen_solver.h
```diff
@@ -162,15 +224,33 @@ namespace oomph
...
-
-    /// Use the eigensolver to find the eigenvalues of a given matrix
-    // void find_eigenvalues(const DoubleMatrixBase &A, const int &n_eval,
-    //                      Vector<std::complex<double> > &eigenvalue,
-    //                      Vector<Vector<double> > &eigenvector);
+                            Vector<std::complex<double>>& alpha,
+                            Vector<double>& beta,
+                            Vector<Vector<std::complex<double>>>& eigenvector)
+    {
+      oomph_info << "Broken, but then don't we want arpack to go anyway?\n";
```
alhazel, line 251: Yes, we should eventually remove ARPACK. Maybe indicate that it is deprecated. Do we have a standard mechanism for that? I can't remember.

src/generic/eigen_solver.h
```diff
@@ -217,37 +297,174 @@ namespace oomph
...
   //=====================================================================
-  /// Class for the LAPACK eigensolver
+  /// Class for the LAPACK QZ eigensolver
   //=====================================================================
   class LAPACK_QZ : public EigenSolver
   {
   public:
     /// Empty constructor
-    LAPACK_QZ() : EigenSolver() {}
+    LAPACK_QZ() : EigenSolver(), Tolerance_for_ccness_check(1.0e-13) {}
```
alhazel, line 311: Given that the Tolerance is only used in PARANOID should this initialisation be PARANOIDed away?

src/generic/eigen_solver.h
```diff
@@ -74,28 +74,92 @@ namespace oomph
...
+    /// eigenvalues are computed), or if the users is happy to deal with NaNs in
+    /// the subsequent post-processing.
+    /// Function is virtual so it can be overloaded for arnoldi type solvers
+    /// that compute the (finite) eigenvalues directly
+    /// At least n_eval eigenvalues are computed.
+    virtual void solve_eigenproblem(
+      Problem* const& problem_pt,
+      const int& n_eval,
+      Vector<std::complex<double>>& eigenvalue,
+      Vector<Vector<std::complex<double>>>& eigenvector)
```
alhazel, line 109: This interface is not suitable for  distributed Problems.. Do we need a complex DoubleVector class?

src/generic/problem.cc
```diff
@@ -8307,13 +8307,125 @@ namespace oomph
...
+  }
+
+  //==================================================================
+  /// Solve the eigenproblem
+  //==================================================================
+  void Problem::solve_eigenproblem(
+    const unsigned& n_eval,
+    Vector<std::complex<double>>& alpha,
+    Vector<double>& beta,
+    Vector<Vector<std::complex<double>>>& eigenvector,
```
alhazel, line 8373: Interface not suitable for distributed problems and therefore not suitable for implementation in the Problem class? ... or should there be a warning?

alhazel (changes requested):
Looks good overall. Moving from DoubleVector to Vector<complex<double> > in the interface may cause problems for future parallel implementations on distributed problems. Of course, there will be other problems there as well, but we may wish to think about this in the interface design. I'm going to request changes, just to force us to have the discussion and so that I learn how this works.

bin/emacs_files_that_contain_string.bash
MatthiasHeilManchester: They should be updated/tracked in case anybody makes any improvements to them. I suppose it's more a question if minor edits (like this one, which I suppose could have been accommodated by passing that argument to the command line) should be committed; this is always the user's decision, to be taken on a case-by-case basis.

MatthiasHeilManchester:
> Looks good overall. Moving from DoubleVector to Vector<complex > in the interface may cause problems for future parallel implementations on distributed problems. Of course, there will be other problems there as well, but we may wish to think about this in the interface design. I'm going to request changes, just to force us to have the discussion and so that I learn how this works.

Yes, that's a good point. Will look into this. I think the "legacy" version is used in some parallel self-tests so I will look into how to update them.

src/generic/eigen_solver.cc
MatthiasHeilManchester: n is the size of the system which is indeed unchanged, the leading dimension follows the pointers to the matrices and is the padded one.

src/generic/eigen_solver.cc
MatthiasHeilManchester: clarity! (Added)

 Done.

src/generic/eigen_solver.cc
MatthiasHeilManchester: It says in the comment above the function that 

 " /// At least n_eval eigenvalues are computed."

This is copied from the (updated) description of the interface in the base class; tends be more of an issue for Arnoldi solvers where one tends specify a smaller number, but the routine may actually return more. The comment is correct here too and I suppose one would expect to receive all n eigenvalues/vectors from a QZ routine.

   Done.

src/generic/eigen_solver.cc
MatthiasHeilManchester: Don't think so, According the lapack's own documentation (copied into the code for clarity) the distinction is made based on complexness, not near-complexness. 

 Done.

src/generic/eigen_solver.cc
MatthiasHeilManchester: Done.

src/generic/eigen_solver.cc
MatthiasHeilManchester: Nothing wrong with what's done (I think!).

  Done.

src/generic/eigen_solver.cc
MatthiasHeilManchester: The commented out oomph-info still labels the logic, so I kept it for that reason, but fair enough. Changed the comment at the very top and removed the else block.

Done.

src/generic/eigen_solver.h
MatthiasHeilManchester: I'll investigate.

src/generic/eigen_solver.h
MatthiasHeilManchester: If you're happy for it to go then we should just make it go. Shall I do the honours?

src/generic/eigen_solver.h
MatthiasHeilManchester: No because that would produce uninitialised data. Baaaaad!.We could paranoid away the definition but that's always a bit scary when people mix and match versions that were compiled with different settings so I'd rather not.

Done.

src/generic/problem.cc
MatthiasHeilManchester: I will make it work for distributed problems.

src/generic/eigen_solver.h
MatthiasHeilManchester: Done.

src/generic/eigen_solver.h
MatthiasHeilManchester: I've made (in the updated pull request) lapack's QZ the default eigensolver but kept the code for arpack around so you can still use it to assess what features of it you need. There are some that are only available in arpack and you may want/need to enable them in the anasazi wrapper. 

Done.

src/generic/eigen_solver.h
MatthiasHeilManchester: Rewrite to DoubleVectors done. The new eigensolvers now return two DoubleVectors, one for the real and one for the imaginary part of the actual, in general complex, eigenvectors.

Done

src/generic/problem.cc
MatthiasHeilManchester: Done.

src/generic/eigen_solver.h
MatthiasHeilManchester: Made lapack QZ the default eigensolver. Kept arpack around so you can revisit it and possibly port any capabilities across to anazsazi. All self-tests pass without it and there are no longer any demo drivers that rely on arpack.

Done.

demo_drivers/bifurcation_tracking/adaptive_hopf_with_separate_meshes.cc
```diff
@@ -1867,7 +1867,7 @@ int main()
   {
    Vector<std::complex<double> > eigenvalues;
    //Now solve the eigenproblem
-   problem.solve_eigenproblem(6,eigenvalues,eigenvectors);
+   problem.solve_eigenproblem_legacy(6,eigenvalues,eigenvectors);
    frequency = eigenvalues[0].imag();
```
alhazel, line 1871: Same comment as above.

src/generic/eigen_solver.h
alhazel: OK. Fair point.

alhazel (commented):
Looks good. I have one query about the _legacy interface and a couple of (trivial) notes.

src/generic/trilinos_eigen_solver.h
```diff
@@ -529,19 +542,23 @@ namespace oomph
       {
         Linear_solver_pt->enable_resolve();
       }
-      // Solve the first vector
 
-      DoubleVector X(x.distribution_pt());
+      // Solve for the first vector (no need for a distribution)
```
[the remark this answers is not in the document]
alhazel: I recall discussions with Richard about efficiency of setting up the distributions, whether it was better to assign the distribution first or not. I can't now recall the details, but that was why the distributions were assigned here. Changing distributions will (obviously) involve some MPI communication. If everything does get set up automatically and efficiently then (obviously) that's fine.

demo_drivers/bifurcation_tracking/adaptive_hopf.cc
```diff
@@ -1304,7 +1304,7 @@ int main()
   {
    Vector<std::complex<double> > eigenvalues;
    //Now solve the eigenproblem
-   problem.solve_eigenproblem(6,eigenvalues,eigenvectors);
+   problem.solve_eigenproblem_legacy(6,eigenvalues,eigenvectors);
```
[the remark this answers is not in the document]
MatthiasHeilManchester: 'fraid so because there was one more interface (the one that only returns the eigenvalues) that can't be distinuished otherwise. It's on your list already: everything that's contains the string _legacy should go.

  Done.

src/generic/trilinos_eigen_solver.h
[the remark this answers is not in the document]
MatthiasHeilManchester: Setting up DoubleVectors (especially in a serial context) always scares me because I can never remember how the distribution thing works (or why I would need to specify it in a serial driver code). Answer: You don't have to and things will be assigned automatically (and generally consistently, using a uniform distribution over all processors). So you can't go wrong with not providing a distribution. Setting them up does not require communication since the distribution is "obvious" once you know how many processors you're running on.

  Done.

src/generic/trilinos_eigen_solver.h
```diff
@@ -629,44 +644,253 @@ namespace oomph
...
+      int verbosity = 0;
+      verbosity += Anasazi::Warnings;
+      // MH has switched off the (overly) verbose output
+      // Could introduce handle to switch it back on.
+      // verbosity+=Anasazi::FinalSummary;
+      // verbosity+=Anasazi::TimingDetails;
+
+
+      Teuchos::ParameterList MyPL;
+      MyPL.set("Which", "LM");
```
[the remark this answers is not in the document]
MatthiasHeilManchester: Added to the issue.

  Done.

MatthiasHeilManchester:
@alhazel All "Done", I think -- feel free to press the green button. Once that's done we need to sort out @PaulAlexander23 's monster pull request (which will be slightly thrashed by this merge. Again. Sorry!). @PaulAlexander23 : Let me know when you've fixed it and/or if you have any conflicts that you can't resolve. And then bully me to get your pull request in before I do anything else (which will be easy because marking hell starts tomorrow.)

alhazel (approved):

PR #84, opened 2022-03-07
Fixes for bubble in channel pull request

PaulAlexander23:
Fix tests in previous pull request that were unintentionally merged in with broken tests.
I had managed to push the fixes these fixes to a wip-branch rather than the bubble-in-channel branch which is why they were missed.
Perhaps we should add a comment in the github help documentation to check that the branch passes all the tests and the best way to do that.


PaulAlexander23:
Merging in to the wrong branch requested!

PR #89, opened 2022-05-15
Add compute_condition_number() to SuperLUSolver and virtual inf_norm() to DoubleMatrixBase

wobblygomboc:
Add helper function compute_condition_number() to SuperLUSolver with two overloads:
- one which assumes the matrix has already been factorised, and so uses the internally stored LU factors to compute the condition number;
- one which accepts a (presumably) unfactorised matrix, performs the factorisation (which then causes the LU factors to be stored internally) and then calls the above overload.

Add virtual double inf_norm() to DoubleMatrixBase with a broken default implementation.

### DIFF
--- a/.github/workflows/self-tests-macos.yaml
+++ b/.github/workflows/self-tests-macos.yaml
@@ -25,12 +25,12 @@ jobs:
       # be unlinked, and thus Python 3 won't be found by other programs. Installing it again doesn't
       # overwrite the Python 2 installation.
       - name: Install Python 2.7
-        uses: LizardByte/setup-python-action@master
+        uses: LizardByte/actions/actions/setup_python@master
         with:
           python-version: "2.7"
 
       - name: Install Python 3.13
-        uses: LizardByte/setup-python-action@master
+        uses: LizardByte/actions/actions/setup_python@master
         with:
           python-version: "3.13"
 
--- a/.github/workflows/self-tests-ubuntu.yaml
+++ b/.github/workflows/self-tests-ubuntu.yaml
@@ -36,12 +36,12 @@ jobs:
       # be unlinked, and thus Python 3 won't be found by other programs. Installing it again doesn't
       # overwrite the Python 2 installation.
       - name: Install Python 2.7
-        uses: LizardByte/setup-python-action@master
+        uses: LizardByte/actions/actions/setup_python@master
         with:
           python-version: "2.7"
 
       - name: Install Python 3.10
-        uses: LizardByte/setup-python-action@master
+        uses: LizardByte/actions/actions/setup_python@master
         with:
           python-version: "3.10"
 


src/generic/linear_solver.h
```diff
@@ -664,6 +663,20 @@ namespace oomph
       Serial_compressed_row_flag = false;
     }
 
+    /// Wrapper to SuperLU's dgscon() which computes the (reciprocal of the)
+    /// condition number of a matrix. This version assumes we're computing the
+    /// condition number of a matrix which has already been factorised and who's
```
MatthiasHeilManchester, line 668: who's --> whose!

src/generic/linear_solver.cc
```diff
@@ -805,6 +805,79 @@ namespace oomph
     }
   } // End of get_total_needed_memory
 
+  //==========================================================================
+  /// Wrapper to SuperLU's dgscon() which computes the (reciprocal of the)
+  /// condition number of a matrix. This version assumes we're computing the
+  /// condition number of a matrix which has already been factorised and who's
```
MatthiasHeilManchester, line 811: who's --> whose

src/generic/matrices.h
```diff
@@ -359,6 +359,15 @@ namespace oomph
     virtual void multiply_transpose(const DoubleVector& x,
                                     DoubleVector& soln) const = 0;
 
+    // compute the infinity-norm
+    virtual double inf_norm() const
+    {
+      throw OomphLibError(
```
MatthiasHeilManchester, line 365: Odd. Is this function implemented for any specific matrix class? Have you forgotten to commit the relevant code? If not there isn't much point committing these changes since nobody would be able to test the code (which I,m sure you have done, somewhere)

MatthiasHeilManchester (changes requested):
See comments

src/generic/matrices.h
wobblygomboc: currently it's only been implemented for CRDoubleMatrix, but I wanted to maintain generality by having compute_condition_number(DoubleMatrixBase* const&) instead of the specific compute_condition_number(CRDoubleMatrix* const&) so that in the future if someone needs this for say a CCDoubleMatrix the function doesn't need to change, they just need to write their own implementation of the inf-norm for a CC matrix.

wobblygomboc: exactly - it already existed just for CRDoubleMatrix, so I've just added a virtual function to the base class so that CRDoubleMatrix::inf_norm() is now an override of this broken DoubleMatrixBase::inf_norm(). And I think this is conceptually correct, since all matrices have an infinity norm so it makes sense to have this functionality in the base class.

Agreed on the self-test.

MatthiasHeilManchester: ...and did you write one?

MatthiasHeilManchester (changes requested):

puneetmatharu:
Has been almost a year since any changes on this PR. What's happening?

MatthiasHeilManchester:
Just going through a Christmas tidy-up operation and stumbled across this. I suppose this got buried under @wobblygomboc 's other tidy-up operations. At the moment they take precedence but we should return to this, i.e. commit all of Christian's changes to the core library before adding his driver codes.

PR #95, opened 2023-01-20
Upgrade autogen based mumps and scalapack

MatthiasHeilManchester:
* mums/scalapack used non-matching parameters in some function calls. From gfortran v 10 upwards this was upgraded from warning to error. Flag to revert back to warning wasn't passed to scalapack demo codes (which we don't want to run anyway). Turns out there's a hidden flag that allows you to switch off the build. Now activated via Makefile.am generated build script and added the relevant fortran compiler flag to all configure option files. 
* added a few extra ifs to avoid building tests when we have mpi but not trilinos.
*  changed the tolerance for checking of ccness in eigensolvers
* tweaked a fpdiff tolerance for macs (Puneet had already fixed that in another not yet pulled-in branch)


puneetmatharu:
LGTM!

PR #100, opened 2023-03-08
Fpdiff.py update

TimoPlath:
Dear Andrew,

I updated fpdiff.py accordingly to our changes in MercuryDPM.
We updated our selftests to also output the directory they are running in at terminal level and this resulted in putting the path+testname as filename. Therefore, fpdiff compared two empty files which lead to a passed selftest. This small fix should now catch if the selftest gold data isn't found and will return a FAILED (return 1) for those tests.

kind regards
Timo

### DIFF
--- a/bin/fpdiff.py
+++ b/bin/fpdiff.py
@@ -80,6 +80,9 @@ def fpdiff_helper(filename1,filename2,relative_error,small,
  # handle the error and so we shouldn't do anything weird here).
  tmpfile1 = read_file(filename1)
  tmpfile2 = read_file(filename2)
+ # filename is passed for the selftest, etc.)
+ if len(tmpfile1) == 0:
+  return 1
 
  #Find the number of lines in each file
  n1 = len(tmpfile1)
--- a/bin/fpdiff.py
+++ b/bin/fpdiff.py
@@ -80,11 +80,12 @@ def fpdiff_helper(filename1,filename2,relative_error,small,
  # handle the error and so we shouldn't do anything weird here).
  tmpfile1 = read_file(filename1)
  tmpfile2 = read_file(filename2)
- # this line catches the error when an empty selftest gold data file is 
+ # this line catches the error when an empty selftest reference data file is 
  # provided and will return a FAIL (e.g. wrong filename is passed for the selftest, etc.)
- if len(tmpfile1) == 0:
-  details_stream.write("\nWarning: gold data file %s is empty and will therefore be treated as a FAIL\n" % filename1)
-  return 1
+ for file_length, filename in [(len(tmpfile1), filename1), (len(tmpfile2), filename2)]:
+  if file_length == 0:
+   details_stream.write(f"\nWarning: file {filename} is empty and the test will therefore be treated as a FAIL\n")
+   return 1
 
  #Find the number of lines in each file
  n1 = len(tmpfile1)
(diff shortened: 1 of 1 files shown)


TimoPlath:
I can somehow not add a reviewer and a label :( 
Hope this still works out for you.

bin/fpdiff.py
```diff
@@ -80,6 +80,10 @@ def fpdiff_helper(filename1,filename2,relative_error,small,
  # handle the error and so we shouldn't do anything weird here).
  tmpfile1 = read_file(filename1)
  tmpfile2 = read_file(filename2)
+ # this line catches the error when an empty selftest gold data file is 
+ # provided and will return a FAIL (e.g. wrong filename is passed for the selftest, etc.)
+ if len(tmpfile1) == 0:
+  return 1
```
puneetmatharu, line 86: I think it would be helpful to provide a meaningful error message rather than silently exiting. Please update accordingly

puneetmatharu (changes requested):
Helloo and congrats on creating the 100th issue/PR to `oomph-lib`! I just have a very small remark to make then I'm happy to merge the PR in

bin/fpdiff.py
alhazel: Agree that it would be useful to output that both files have zero length, which is treated as a fail.

TimoPlath: that makes sense, newest added an error message to the details_stream

bin/fpdiff.py
```diff
@@ -83,6 +83,7 @@ def fpdiff_helper(filename1,filename2,relative_error,small,
  # this line catches the error when an empty selftest gold data file is 
  # provided and will return a FAIL (e.g. wrong filename is passed for the selftest, etc.)
  if len(tmpfile1) == 0:
+  details_stream.write("\nWarning: gold data file %s is empty and will therefore be treated as a FAIL\n" % filename1)
```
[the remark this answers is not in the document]
TimoPlath: good idea, I rephrased it to the following:

```
 # this line catches the error when an empty selftest reference data file is 
 # provided and will return a FAIL (e.g. wrong filename is passed for the selftest, etc.)
 for file_length, filename in [(len(tmpfile1), filename1), (len(tmpfile2), filename2)]:
  if file_length == 0:
   details_stream.write(f"\nWarning: file {filename} is empty and the test will therefore be treated as a FAIL\n")
   return 1
```
regarding your 2nd point: For me the indentation is 1 space throughout the whole file, so I kept it now like that.

bin/fpdiff.py
```diff
@@ -80,6 +80,12 @@ def fpdiff_helper(filename1,filename2,relative_error,small,
  # handle the error and so we shouldn't do anything weird here).
  tmpfile1 = read_file(filename1)
  tmpfile2 = read_file(filename2)
+ # this line catches the error when an empty selftest reference data file is 
+ # provided and will return a FAIL (e.g. wrong filename is passed for the selftest, etc.)
```
puneetmatharu, line 84: Ah, one final thing: the comment isn't entirely correct -- if the wrong filename is passed, `readfile(...)` will try to open a file that doesn't exist and will throw an error. In this case, the user won't even reach this line. If you remove the text "(e.g. wrong filename is passed for the selftest, etc.)", I'm happy to merge this in

TimoPlath: done

puneetmatharu:
LGTM!

TimoPlath:
Great thanks for merging!

PR #101, opened 2023-06-19
matthias office machine private config options

MatthiasHeilManchester:


MatthiasHeilManchester:
merged myself since it's just two private config files; the changes to cgal (no testing) appear to have been checked in via some other route already

PR #103, opened 2023-10-09
Update search_results.txt to point to new CSE.

PaulAlexander23:
Change the cx variable, which points to which Programmable Search Engine is being used, to a new one for the github version of the oomph-lib documentation.
I currently have ownership of the new Programmable Search Engine but can add/transfer ownership as needed.

### DIFF
--- a/doc/search_results/search_results.txt
+++ b/doc/search_results/search_results.txt
@@ -17,7 +17,7 @@ document.getElementById("search_warning").style.display = "block";
 
 <script>
   (function() {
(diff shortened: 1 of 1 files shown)


PaulAlexander23:
Alternatively, the original Programmable Search Engine could be updated. However I don't know who the owner is to contact them. This would mean that no commit is required for the github.

MatthiasHeilManchester:
Brilliant. I think I owned the old one but can't remember anything about it. Let's discuss how to share access to the new search engine at the next oomph-lunch.

PR #106, opened 2023-12-04
Fix search

MatthiasHeilManchester:
Some minor tidy to Paul's search machinery. Builds cleanly when commit to the main branch in my own repository so should be safe to move across.

### DIFF
--- a/bin/wrapper_for_validate.sh
+++ b/bin/wrapper_for_validate.sh
@@ -29,6 +29,23 @@ echo " "
 cat validate_sh_timing_generated_by_make_check.dat
 echo " "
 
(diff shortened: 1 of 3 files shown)


puneetmatharu:
No objections here! Merging it in.

puneetmatharu:
Hmm. Why are you trying to merge directly into `main` rather than `development`?

MatthiasHeilManchester:
> Hmm. Why are you trying to merge directly into `main` rather than `development`?

True. There was some logic in the sense that the change only affects functionality that matters in the main branch (and only ever on the github hosted webpages) but I'll change it. Should then pull the changes across to main afterwards.

PR #107, opened 2023-12-12
Add mesh copy/build from an existing mesh.

PaulAlexander23:
Add a new function to build a new mesh with new elements from an existing one.
New nodes and elements are created in place of the old ones, where the old elements need to have the same pattern of nodes.
The boundary information is set up after the all the nodes and elements have been created.

Please advise as to whether this should be further up the inheritance ladder (currently in TriangleMesh to have access to the template element), and what additional checks would be necessary.

Thanks.


puneetmatharu:
Hi, I'm limited on time to look on this so I'll be blunt: I don't think this should be merged in. However, @MatthiasHeilManchester may wish to override my suggestion, so please be patient and wait for him to reply.

Even at a quick glance, `build_from_another_mesh(...)` is ambiguous and it's not entirely clear what its purpose is. Unless I am mistaken, you simply want a "duplicate_me(...)" factory function for a mesh; the current API you have provided:

```cpp
void TriangleMesh<ELEMENT>::build_from_another_mesh(Mesh* orig_mesh_pt)
```

does not make this clear; it implies that you want to take any arbitrary mesh and try to create a `TriangleMesh` from it. What happens if I pass in a `QuadMesh`? (It will break.) Not to mention, meshes derived from `TriangleMesh` (and any other `Mesh`) can contain arbitrarily complex information that you may not copy across with this implementation, forcing users to write their own custom version for any other mesh. These sort of things are hard to implement correctly and would typically involve adding a broken virtual function to every mesh that must be implemented for each mesh; I think this an unnecessary amount of work for such a simple objective. Not to mention, it's rare that you'll want to make copies of meshes, since they are typically expensive to generate/store in memory.

Admittedly, the copy functionality seems a bit overengineered. I haven't ever encountered this use-case before. In the demo driver you provided, I would just take the pointer you passed to the `Problem` object and use that as my mesh instead of copying it (which is, again, expensive). If I really needed to create a new problem at a later point in time from an old one, I would dump a copy of my mesh to disk then load it as and when I need. By doing so, I avoid holding two copies of a `Problem` object or mesh in main memory.

If you *really*, *really* need this functionality, why not just create your own derived class for the mesh you need, add the extra copy functionality and use that?

PaulAlexander23:
All valid points.
I need this functionality to create a new problem with new elements on a previous mesh. (for performing a linear stability on top of a nonlinear steady state.)
I did originally create a new class called MyTriangleMesh, and that works great for my problem.
I mentioned this in oomph-lunch and Matthias suggested that this might be useful in other cases also, hence the pull request.
This is definitely work in progress to clean it up. I don't know a better workflow to get people opinions before submitting a pull request.
Name suggestions welcome. Originally we had deep_copy, but that is not what the function does really.
I'll update the interface and add some more checks after Matthias has had a look.

puneetmatharu:
> All valid points.
I need this functionality to create a new problem with new elements on a previous mesh. (for performing a linear stability on top of a nonlinear steady state.)
I did originally create a new class called MyTriangleMesh, and that works great for my problem.

I see, that's good to know.

> I mentioned this in oomph-lunch and Matthias suggested that this might be useful in other cases also, hence the pull request.

In that case, there's possible merit in adding this feature in.

> This is definitely work in progress to clean it up. I don't know a better workflow to get people opinions before submitting a pull request.

I would recommend tidying it up as much as possible before submitting a PR then when you're happy to propose the change, create the PR and we can just provide proposals. It's probably the easiest route. (Likely what you have already done.)

> Name suggestions welcome. Originally we had deep_copy, but that is not what the function does really.

Hmm, after another look, maybe I have misunderstood and that you're not planning to do a deep copy. From what I can see you're just copying the spatial data across. What about everything else a mesh might contain?

Also, you're resizing the `Element_pt` and `Node_pt` without checking that they're empty. If they have already been populated and you just overwrite the pointers, you'll get a memory leak. 

> I'll update the interface and add some more checks after Matthias has had a look.

Got it.

PaulAlexander23:
Pulled in latest changes to development.
Please check the names of functions, variables etc. If there are too many changes, then we can put this pull request into a draft state to finish up later and pull in the development -> main first.

src/meshes/triangle_mesh.template.cc
```diff
@@ -36,6 +36,177 @@
...
+  //======================================================================
+  /// Build mesh from the structure and geometry of another mesh.
+  /// A lot of this code is repeated in the construction of the mesh from
+  /// the scaffold mesh.
+  //======================================================================
+  template<class ELEMENT>
+  template<class ORIG_ELEMENT>
+  void TriangleMesh<ELEMENT>::build_from_another_mesh(
+    TriangleMesh<ORIG_ELEMENT>* orig_mesh_pt)
+  {
```
MatthiasHeilManchester, line 48: Didn't we need a warning that this won't work if run in parallel (if true!). Maybe I mis-remembered

MatthiasHeilManchester (changes requested):
extra issue that I'd forgotten.

src/meshes/triangle_mesh.template.cc
PaulAlexander23: It currently passes the MPI tests on Github but probably won't work if the mesh is segregated etc.
What should I be testing for? A blank MPI test seems a bit excessive.

demo_drivers/meshing/Makefile.am
```diff
@@ -8,6 +8,7 @@ mesh_from_xfig_triangle \
 mesh_from_vmtk \
 quad_from_triangle_mesh \
 quadratic_vmtk_tetgen \
+mesh_from_another_mesh_poisson \
```
MatthiasHeilManchester, line 11: triangle_mesh_from_another_triangle_mesh. Can save some space (just in the directory name) by omitting poisson.

src/meshes/triangle_mesh.template.cc
```diff
@@ -36,6 +36,189 @@
...
 {
+  //======================================================================
+  /// Build mesh from the structure and geometry of another mesh but
+  /// replace element by a different type (passed in via the template
+  /// parameter as usual).
+  /// A lot of this code is repeated in the construction of the mesh from
+  /// the scaffold mesh.
+  //======================================================================
+  template<class ELEMENT>
+  template<class ORIG_ELEMENT>
```
MatthiasHeilManchester, line 47: I think there was a query if we needed the second template at all.

src/meshes/triangle_mesh.template.cc
MatthiasHeilManchester: Ah, yes, I mean distributed. I think meshes have a boolean member function that says if they're distributed, so if you suspect that this functionality won't work in such cases (I agree) we should bail out as a precaution.

src/meshes/triangle_mesh.template.cc
MatthiasHeilManchester, line 47: Do we need the second template arg at all? Orig mesh can simply be a Triangle base mesh.

MatthiasHeilManchester (changes requested):

src/meshes/triangle_mesh.template.cc
MatthiasHeilManchester: Mesh::finite_element_pt(...) does the cast automatically (and dies if it fails).

PaulAlexander23: I've removed the second template and added FiniteElement instead however in testing I get a new mesh structure!?
I'm having to do a dynamic cast from my old mesh type to TriangleMesh<FiniteElement>* which might be where the issue originates from. Keeping everything in its original state was why I was double templating.

This is probably more complicated than it needs to be.
I suggest that I close this pull request for now, and we focus on the other one until the purpose of this one is clearer.

MatthiasHeilManchester:
probably easiest to have a quick chat, but for now I'll pull the development branch across to main and then we (you in particular!) deserve biiiiig beer.

 m.

Sent from my iPad

Prof. Matthias Heil
Alan Turing Building, Room 2.224
School of Mathematics
University of Manchester
Oxford Road<x-apple-data-detectors://4>
Manchester M13 9PL<x-apple-data-detectors://4>
U.K.<x-apple-data-detectors://4>

Tel. +44 (0)161 275 5808<tel:+44%20161%20275%205808>
email: ***@***.******@***.***>
           ***@***.******@***.***>

WWW: https://personalpages.manchester.ac.uk/staff/matthias.heil/
             https://www.matthias-heil.co.uk<https://www.matthias-heil.co.uk/>
             https://www.matthias-heil.co.uk/rvy_racing

Editor in Chief of Journal of Fluids and Structures
Associate Editor of Journal of Engineering Mathematics

NEWS:   Release 2.0 of oomph-lib, the object-oriented
       multi-physics finite-element library is now available
       as free open-source software at

           http://www.oomph-lib.org<http://www.oomph-lib.org/>


On 17 Dec 2024, at 17:11, PaulAlexander23 ***@***.***> wrote:

﻿

@PaulAlexander23 commented on this pull request.

________________________________

In src/meshes/triangle_mesh.template.cc<https://github.com/oomph-lib/oomph-lib/pull/107#discussion_r1888908978>:

> @@ -36,6 +36,189 @@

 namespace oomph
 {
+  //======================================================================
+  /// Build mesh from the structure and geometry of another mesh but
+  /// replace element by a different type (passed in via the template
+  /// parameter as usual).
+  /// A lot of this code is repeated in the construction of the mesh from
+  /// the scaffold mesh.
+  //======================================================================
+  template<class ELEMENT>
+  template<class ORIG_ELEMENT>


I've removed the second template and added FiniteElement instead however in testing I get a new mesh structure!?
I'm having to do a dynamic cast from my old mesh type to TriangleMesh* which might be where the issue originates from. Keeping everything in its original state was why I was double templating.

This is probably more complicated than it needs to be.
I suggest that I close this pull request for now, and we focus on the other one until the purpose of this one is clearer.

PR #108, opened 2023-12-13
Fix and warn for TGauss<2,16> scheme

AidanRetallick:
Corrected integration weights for TGauss<2,16> (which were out by a factor of 4) so that they now give a unit triangle area of 0.5 rather than 2.

Also, because this scheme is slightly unconventional in that some of is knots are external, and weights negative, a static flag has been added which triggers a warning if built with PARANOID to explain this to the user exactly once if used.


puneetmatharu:
Thanks for the PR! As mentioned on your other PR, **please remember to share links to your self-tests to demonstrate that you have not introduced breaking changes when creating a PR**.

@MatthiasHeilManchester I will leave it to you to sign this off as I'm not familiar with this code.

src/generic/integral.h
```diff
@@ -1055,9 +1055,35 @@ namespace oomph
     /// Array to hold the weights and knots (defined in cc file)
     static const double Knot[52][2], Weight[52];
 
+#ifdef PARANOID
+    /// A flag to track whether we have warned the user about the exterior knots
+    /// and negative weights in this scheme.
+    static bool User_has_been_warned;
```
MatthiasHeilManchester, line 1061: Maybe I'm blind but where is this initialised to false?

src/generic/integral.cc
```diff
@@ -551,26 +551,26 @@ namespace oomph
     {0.08528361568265699, 0.014317320230681},
     {0.900399064086661, 0.08528361568265699},
     {0.014317320230681, 0.900399064086661}};
-  // Modified by DR 2017 to include factor 1/2
+  // Modified by AR 2023 to include factor 1/2
```
MatthiasHeilManchester, line 554: @puneetmatharu is right (in his general comment on this pull request). We should briefly mention where this scheme is used (in the existing self-tests). Possibly not at all (yet -- until your FvK stuff is added) in which case, the relevant driver code may want to include a little stand-alone test that integrates the area of the domain -- this would be off by a factor of [dunno -- non-one] if the correction was wrong.

MatthiasHeilManchester (changes requested):
Embedded two queries

src/generic/integral.cc
AidanRetallick: Not used by anything (yet), should I put it in the comments?

src/generic/integral.h
AidanRetallick: Woops! I've just forgotten to commit that change. It is in the .cc file (I'll commit it now).

MatthiasHeilManchester (approved):

PR #109, opened 2023-12-13
Fix error estimator

AidanRetallick:
Change `Z2ErrorEstimator::doc_flux` file naming procedure.

Previously, if oomph-lib was compiled with MPI (either for distribution or to build MUMPS) then `Z2ErrorEstimator` expected `comm_pt` to be set. If, for example, you wanted to build MUMPS but not distribute your problem you would run into a segfault when trying to dereference a null pt.

Note now also that if the problem is not distributed, the filenames are:
> flux_rec.dat
> flux_fe.dat

rather than 
> flux_rec_on_proc_0.dat
> flux_fe_on_proc_0.dat


puneetmatharu:
All looks pretty straightforward. I have no objections. **However, please remember to share links to your self-tests to demonstrate that your addition doesn't introduce breaking changes.** For now I have looked at your Actions and it seems like the only failure is a somewhat-expected one on macOS. Thanks!

PR #110, opened 2023-12-18
Add adaptive dt safety factor

AidanRetallick:
Adding safety factor to the adaptive time step estimator. Allows us to aim for an error strictly less than our max tolerance and return appropriate size steps.

New variable `Target_error_safety_factor` defines the target error, TAR, in terms of the tolerance, TOL:
> TAR = `Target_error_safety_factor` * TOL

And is accessed using the function `target_error_safety_factor()`.

By default this new variable is set to 1.0 so as to not change oomph-libs default behaviour at this time.

**[Ubuntu tests](https://github.com/AidanRetallick/oomph-lib/actions/runs/7209926347)** (passing)
**[Mac tests](https://github.com/AidanRetallick/oomph-lib/actions/runs/7209926325)** (passing)


src/generic/problem.cc
```diff
@@ -11431,11 +11434,20 @@ namespace oomph
...
           }
           oomph_info
-            << "Note: This behaviour can be adjusted by changing the protected "
+            << "Note: This behaviour can be adjusted by changing the protected"
             << "boolean" << std::endl
+            << "    Problem::Keep_temporal_error_below_tolerance" << std::endl
             << std::endl
-            << "    Problem::Keep_temporal_error_below_tolerance" << std::endl;
+            << "Also, if you are noticing that many of your timesteps result"
+            << "in error > tolerance, try reducing the target error with"
```
MatthiasHeilManchester, line 11442: ...should this be '...try reducing the "target error safety factor from its default value of 1.0. This will reduce the next timestep below the default estimate which aims to make the error close to the target error. This default strategy makes the timestep as large as possible while respecting the requested tolerance, but risks that the next timestep is rejected if the error estimate turns out to be too conservative. Setting the safety factor to too small a value will make the timesteps unnecessarily small; choosing a value that is too large may lead to an unnecessarily large number of rejections, forcing a recomputation of the solution at a reduced  timestep. Neither is optimal; a problem dependent compromise is needed; see ***ref**** for details.'

src/generic/problem.h
```diff
@@ -717,6 +717,22 @@ namespace oomph
     /// with a smaller dt).
     double DTSF_min_decrease;
 
+    /// Safety factor to ensure we are aiming for a target error, TARGET,
+    /// that is below our tolerance:
+    ///     TARGET = Target_error_safety_factor * TOL
+    /// For this to make sense Target_error_safety_factor should be <1.0. If
+    /// Keep_temporal_error_below_tolerance is set to true (default) then,
+    /// without this, timesteps produced by the adaptive time-step suggester
```
MatthiasHeilManchester, line 725: ..timesteps suggested by the adaptive timestepper...

src/generic/problem.h
```diff
@@ -717,6 +717,22 @@ namespace oomph
...
+    /// can be expected to fail (exceed TOL) about half of the time.
```
MatthiasHeilManchester, line 726: ...be expected to lead to rejection (because the error exceeds TOL)...

MatthiasHeilManchester (changes requested):
Embedded two suggestions for rephrasing comments

src/generic/problem.cc
AidanRetallick: I'll add most of what you've said but, importantly, I think the part about
 "...the default estimate which aims to make the error close to the **target** error." 
should be
 "...the default estimate which aims to make the error close to the **tolerance** error."

The idea of the safety factor is that it is changing the target error to be something less than tolerance (provided it is <1.0) whereas the current strategy is to try and produce an error that is equal to our tolerance (which results in **A LOT** of rejected steps *something I want to talk about in oomph-lib* (again)).

However, I agree it is important to clarify the limiting behaviour of too small vs too big in this messge.

src/generic/problem.h
AidanRetallick: Yep, correct. Will change.

src/generic/problem.h
AidanRetallick: Yep. Will change.

PR #111, opened 2023-12-21
Fix axisym fvk stress

AidanRetallick:
Add stress calculation at origin of the axisymmetric domain fvk elements.

Previously, if we attempted to retrieve the stress at r=0 for either the displacement based or airy stress axisymmetric foeppl von karman elements, the respective `interpolated_stress(...)` function would give up and return false. This is because the standard analytical form for the stress in both cases contains a u_r/r term. There is however no stress singularity and making use of connectivity at the origin (u_r=0) we get u_r/r -> d2u/dr2 as r->0 allows us to retrieve the real stress without giving up.

Each set of equations -- displacement based and airy stress -- has had its stress function altered such that it uses an alternate form of stress at the origin rather than giving up. As such, there is no need for a bool return and the functions `interpolated_stress(...)` are now of void type and the code that interfaces with these functions no longer expects the bool return. Solutions should generaly be unaffected as there are no Gauss points at the origin however the origin can now be output after solving meaning the relevant test datasets have been updated to include a single extra datapoint. I have attached some ugly plots of the r and theta stresses near the origin in both the old and new test datasets to show they differ only by a plot point at r=0 (*but I stress that they are ugly*).

Lastly, in both versions the entirety of `compute_error(...)` body was repeated causing the function to (without any other side effects) be executed twice. This has been corrected.

[Ubuntu self-tests](https://github.com/AidanRetallick/oomph-lib/actions/runs/7276617975) (passing)
[Mac self-tests](https://github.com/AidanRetallick/oomph-lib/actions/runs/7276617967) (passing)

disp_based_axysym_stress_comparison_zoom
axysym_stress_comparison_zoom

### DIFF
--- a/src/axisym_displ_based_foeppl_von_karman/axisym_displ_based_fvk_elements.cc
+++ b/src/axisym_displ_based_foeppl_von_karman/axisym_displ_based_fvk_elements.cc
@@ -286,8 +286,6 @@ namespace oomph
       // Note that this process results in s_{rr}=s_{\phi\phi} at the origin
       // which is expected as the two coordinate directions are
       // indistinguishable at the origin under axisymmetry.
-      // [zdec] Should this be interpolated_r < tolerance to prevent floating
-      // point comparison? -- AidanR
       if (interpolated_r == 0.0)
       {
         // Compute the stresses:
--- a/src/axisym_foeppl_von_karman/axisym_fvk_elements.cc
+++ b/src/axisym_foeppl_von_karman/axisym_fvk_elements.cc
@@ -316,8 +316,6 @@ namespace oomph
       } // End of loop over nodes
 
       // Compute stresses
-      // [zdec] Should this be interpolated_r < tolerance to prevent floating
-      // point comparison? -- AidanR
       if (interpolated_r == 0.0)
       {
         // To avoid dividing by zero we take advantage of axisymmetry at the


src/axisym_displ_based_foeppl_von_karman/axisym_displ_based_fvk_elements.cc
```diff
@@ -289,23 +276,40 @@ namespace oomph
...
+      // dividing by zero.  It can be found from the form below by assuming:
+      // a) u_r=0 at the origin (required under axisymmetry unless there is a
+      // puncture) which leads to the term u_r/r -> du_r/dr as r -> 0 (by
+      // L'Hopital's rule), and
+      // b) given dw/dr=0 at the origin (required for axisymmetry as there can
+      // be no kink in a plate).
+      // Note that this process results in s_{rr}=s_{\phi\phi} at the origin
+      // which is expected as the two coordinate directions are
+      // indistinguishable at the origin under axisymmetry.
+      // [zdec] Should this be interpolated_r < tolerance to prevent floating
```
MatthiasHeilManchester, line 289: I think it's fine; the error really only occurs when r is exactly zero. Just change comment to:
"Yes, this really does require a floating point comparison since the division by zero only occurs when the radius is exactly zero."

src/axisym_foeppl_von_karman/axisym_fvk_elements.cc
```diff
@@ -331,12 +316,20 @@ namespace oomph
       } // End of loop over nodes
 
       // Compute stresses
-      sigma_r_r = interpolated_dphi_dr / interpolated_r;
+      // [zdec] Should this be interpolated_r < tolerance to prevent floating
```
MatthiasHeilManchester, line 319: Ditto (see other comment)

src/axisym_displ_based_foeppl_von_karman/axisym_displ_based_fvk_elements.cc
```diff
@@ -493,66 +496,6 @@ namespace oomph
       norm += exact_soln[0] * exact_soln[0] * W;
       error += (exact_soln[0] - w_fe) * (exact_soln[0] - w_fe) * W;
     }
-
```
MatthiasHeilManchester, line 496: Mention in commit that duplicate code was removed.

src/axisym_foeppl_von_karman/axisym_fvk_elements.cc
```diff
@@ -525,66 +517,6 @@ namespace oomph
       norm += exact_soln[0] * exact_soln[0] * W;
       error += (exact_soln[0] - w_fe) * (exact_soln[0] - w_fe) * W;
     }
-
-    {
```
MatthiasHeilManchester, line 529: Mention in commit that duplicate code was removed.

MatthiasHeilManchester (changes requested):
Minor tweak requests embedded.

src/axisym_foeppl_von_karman/axisym_fvk_elements.cc
AidanRetallick: Now mentioned in the commit which made this change

src/axisym_displ_based_foeppl_von_karman/axisym_displ_based_fvk_elements.cc
AidanRetallick: Now mentioned in the commit which made this change

AidanRetallick:
I've rebased onto the latest development commit: in the process, updating the commit that removed the repeated code in order to mention this in its commit message.

Also the developer comments noting whether floating point comparisons should be managed differently have been removed.

MatthiasHeilManchester (approved):

PR #113, opened 2023-12-28
Fix heisenbug dec 2023

MatthiasHeilManchester:
- Work-around for bizarre MPI problem in distributed triangle meshes. Code used to 
seg fault when going into MPI_Wait if  oomph-lib was compiled with any 
optimisation, i.e. -O1 upwards.   ddd-based investigations unclear. Trial and error showed that
 passing a new (rather than re-used) MPI_Request object to that
 call (and the associated collective communications that it's
 waiting for) fixed the problem. Somewhat unsatisfactory and definitely
 one to keep an eye on...

ANDREW: Can you please clone my branch and check if it works on your machine too (either pre- or post-accept for this merge)

- A few more little tidies and modified the build machinery to explicity suppress Qt5-related builds in CGAL.

- Ubuntu passes cleanly

     https://github.com/MatthiasHeilManchester/oomph-lib/actions/runs/7346957076

- macos still has the usual failure in the elastic bretherton code

     https://github.com/MatthiasHeilManchester/oomph-lib/actions/runs/7346957077

### DIFF
--- a/src/meshes/triangle_mesh.template.cc
+++ b/src/meshes/triangle_mesh.template.cc
@@ -6213,6 +6213,12 @@ namespace oomph
         }
 
         // --------------
+
+
+        
+         // Work-around for odd seg fault when calling MPI_Wait below.
+         MPI_Request my_request;
+         
         // Double data
         unsigned nflat_double_send = flat_double_send_packed_data.size();
         MPI_Isend(&nflat_double_send,
@@ -6221,8 +6227,8 @@ namespace oomph
                   send_proc,
                   3,
                   comm_pt->mpi_comm(),
-                  &request);
-
+                  &my_request);
+        
         unsigned nflat_double_receive = 0;
         MPI_Recv(&nflat_double_receive,
                  1,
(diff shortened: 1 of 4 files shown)


src/meshes/triangle_mesh.template.cc
```diff
@@ -34560,10 +34566,6 @@ namespace oomph
         this->copy_connection_information(polygon_pt->polyline_pt(p),
                                           tmp_curve_section_pt);
 
-        // Now update the polyline according to the new vertices but
```
alhazel, line 34563: Is this now redundant? Looks a bit dangerous to remove, unless the potential memory leak is addressed elsewhere.

alhazel:
I have confirmed that the branch works on my machine under both OpenMPI and mpich. Flags in the validate script need to be changed for Mpich are to allow --output-filename (OpenMPI) and --outfile-pattern (mpich).  Once this change is included then I'm happy to approve the request.

src/meshes/triangle_mesh.template.cc
MatthiasHeilManchester: not yet! I've just pushed the change to the mpirun flag. May as well wait until the github ci tests pass for that too.

MatthiasHeilManchester:
> I have confirmed that the branch works on my machine under both OpenMPI and mpich. Flags in the validate script need to be changed for Mpich are to allow --output-filename (OpenMPI) and --outfile-pattern (mpich). Once this change is included then I'm happy to approve the request.

Done. All self tests pass on github:

- https://github.com/MatthiasHeilManchester/oomph-lib/actions/runs/7643357264
- https://github.com/MatthiasHeilManchester/oomph-lib/actions/runs/7643357256

src/meshes/triangle_mesh.template.cc
alhazel: OK. All Approved.

PR #115, opened 2024-01-05
Add barycentric element splitting to triangle mesh

AidanRetallick:
Summary
-------------
Similar to the tet element splitting which is already implemented, this PR will add two functions to the TriangleMeshBase class which can be used to alter a mesh such that it has no elements with more than one boundary edge. The specific motivation being the curvable triangular Bell elements (not yet in oomph-lib proper) which can only be upgraded to have one curved boundary edge. The process is only implemented for NNODE_1D=2,3,4 (as most things in oomph-lib are), below is an example of the process for an NNODE_1D=3 mesh.

corner_split_nnode1d_3_trimmed

Changes
-------------
Two nwo functions have been added to TriangleMeshBase.

The first function takes a pointer to an element and will split that element into three by bisecting the triangles vertices to the elements centroid. It stores pointers to the new elements and their nodes in vector containers which are also provided as arguments:

    template<class ELEMENT>
    void split_element_through_centroid(
      FiniteElement* const& el_pt,
      Vector<FiniteElement*>& new_el_pt,
      Vector<Node*>& new_nod_pt,
      TimeStepper* const& time_stepper_pt = &Mesh::Default_TimeStepper);

The second function will locate all elements with at least two edges on boundaries (either the same or different boundary) and split all such elements using the above function  -- also doing the relevant bookkeeping to update the mesh data -- resulting in a mesh where no element has more than one boundary edge:

    template<class ELEMENT>
    void split_elements_with_multiple_boundary_edges(
      TimeStepper* const& time_stepper_pt = &Mesh::Default_TimeStepper);

Both functions optionally accept a pointer to a time stepper which will in turn be passed onto the new nodes.

Motivation
-------------
The motivation for this is the curvable triangular elements that I am working on, which can be upgraded to interpolate a curved boundary using a specialised basis. However, this basis is only designed to interpolate triangles with one curved edge, hence if there were a corner element straddling two boundaries for example, it will not be able to interpolate both properly. By splitting all such elements through their vertices as so:
    
        //  o2                          o2              //
        //  | \                         |\\             //
        //  |   \                       | \ \           //
        //  |     \          becomes:   |  \  \         //
        //  |       \                   |   c_  \       //
        //  |         \                 | /    - _\     //
        //  o0----------o1              o0----------o1  //

all elements are guarenteed to have at most one boundary edge.

Notes
-------------
- This pull request is just a draft as there will definitely be more work required before it is mergable. I'm making it to allow Matthias (and whoever might be interested) to snoop and tell me off for anything I may have missed.

- The overall structure of the splitting process is similar to the tet version of this process for splitting corner tets however I have extracted the part of the code that does the splitting, partly because I think it is nicer code, but also because it is possible someone may want to use it to split elements in some other context. Additionally, this splits all elements with multiple boundary edges (even two edges on the same boundary) whereas I believe the tet version is just in corners.

- A negative consequence of splitting in this manner is that it results in slenderer elements than originally in the mesh as angles are *approximately* bisected.

- TODO:
  - A self test needs writing before this is merged.
  - There is some lookup scheme bookkeeping still missing from the end of `split_elements_with_multiple_boundary_edges`. 
  - Almost definitely needs some Paranoid checks to ensure these functions are more robust and not being misused (ideas welcome).
  - Currently there is nothing ensuring that the element pointer belongs to the mesh doing the splitting or that it is even a triangle element. This could lead to problems. Maybe a Paranoid check for this will be enough.
  - Remove comments-to-self and debug outputs.


puneetmatharu:
This is 1.5y old now. What's the plan here? @MatthiasHeilManchester

MatthiasHeilManchester:
> This is 1.5y old now. What's the plan here? @MatthiasHeilManchester

Aidan's stuff is still being cleaned up. We're actually actively working on it, so this will go in with the rest of his stuff. So let's keep it alive. It'll turn into something "well good"! @AidanRetallick

PR #118, opened 2024-02-02
Remove arpack from the library

alhazel:
This pull request simply removes all traces of ARPACK from oomph-lib. Well, the word is left in the custom dictionary, and in the archive commit logs, but otherwise it's gone.


MatthiasHeilManchester:
Brilliant. That looks extremely comprehensive; hadn't thought of some of the places where arpack was being referred to. Let's hear it for find/grep (I assume!)

PR #119, opened 2024-02-14
Fix a bug in error handling

PaulAlexander23:
- Make sure that all the errors inherit from the OomphLibError.
- Make member data private within the error classes.
- Add an inverted element error to allow the adaptive time stepper and arc length continuation to continue, with reduced step size, rather than terminating.


src/generic/problem.cc
```diff
@@ -10911,12 +10911,22 @@ namespace oomph
...
+                       << std::endl;
             STEP_REJECTED = true;
             // Let's take a smaller step
             Ds_current *= (2.0 / 3.0);
           }
         }
+        catch (InvertedElementError const& error)
+        {
+          oomph_info
+            << "STEP REJECTED DUE TO INVENTED ELEMENTS --- TRYING AGAIN"
```
MatthiasHeilManchester, line 10924: invented or inverted? :)

src/generic/problem.h
```diff
@@ -3100,9 +3103,9 @@ namespace oomph
   //=======================================================================
   /// A class to handle errors in the Newton solver
   //=======================================================================
-  class NewtonSolverError
+  class NewtonSolverError : public OomphLibError
   {
-  public:
+  private:
     /// Error in the linear solver
     bool linear_solver_error;
```
MatthiasHeilManchester, line 3110: Uppercase!

src/generic/problem.h
```diff
@@ -3112,24 +3115,58 @@ namespace oomph
...
 
+  public:
     /// Default constructor, does nothing
-    NewtonSolverError() : linear_solver_error(false), iterations(0), maxres(0.0)
+    NewtonSolverError()
+      : OomphLibError("There was an error in the Newton Solver.",
+                      OOMPH_CURRENT_FUNCTION,
+                      OOMPH_EXCEPTION_LOCATION),
+        linear_solver_error(false),
+        iterations(0),
```
MatthiasHeilManchester, line 3125: "iterations" uppercase too. Not sure where it's defined, actually.

MatthiasHeilManchester (changes requested):
Couple of minor change requests. Given that @alhazel wrote most of this error handling stuff I wouldn't mind if he could have a quick look through this too and do the ultimate accept.

src/generic/problem.h
PaulAlexander23: Fixed!

src/generic/problem.h
PaulAlexander23: Fixed!

src/generic/problem.cc
PaulAlexander23: Good spot. It should be inverted.

src/multi_physics/segregated_fsi_solver.cc
```diff
@@ -942,11 +958,19 @@ namespace oomph
...
+              true);
           }
           else
           {
-            throw SegregatedSolverError(false);
+            throw SegregatedSolverError(
+              "Error occured in Segregated solver. \n",
+              OOMPH_CURRENT_FUNCTION,
+              OOMPH_EXCEPTION_LOCATION,
+              false);
```
PaulAlexander23, line 973: I don't like the flag in the error arguments. I think we should throw two different errors instead.

src/generic/problem.h
```diff
@@ -3100,35 +3103,69 @@ namespace oomph
   //=======================================================================
   /// A class to handle errors in the Newton solver
   //=======================================================================
-  class NewtonSolverError
+  class NewtonSolverError : public OomphLibError
   {
-  public:
+  private:
```
PaulAlexander23, line 3108: If we make the public variables private, then we are changing the interface of the class. I've added getter functions but if someone else's code tries to access the class's variable then it'll now throw an error.

MatthiasHeilManchester: If they read public data they deserve everything they get. Just make sure there's an explicit comment in the commit log.

MatthiasHeilManchester (approved):

src/generic/problem.h
alhazel: I disagree with this broad statement that public data should not be read. The whole point of error objects is that they are essentially structures with a number of diagnostic variables that should, in my view, be public rather than going through the overhead of set and get functions. Set should never be needed because the only time variables are set is during creation. The object is created, thrown and destroyed after being caught or rethrown, so there is no potential for unexpected side effects. No benefit to data encapsulation in my view.

src/generic/problem.h
```diff
@@ -3050,35 +3053,69 @@ namespace oomph
...
+                      OOMPH_CURRENT_FUNCTION,
+                      OOMPH_EXCEPTION_LOCATION),
+        Linear_solver_error(false),
+        Iterations(Passed_iterations),
+        Maxres(Passed_maxres)
+    {
+    }
+
+    /// Access function to the error in the linear solver
+    bool get_linear_solver_error()
```
MatthiasHeilManchester, line 3103: Can we kill the "get_" not really needed (and not typically done in oomph-lib)

src/generic/problem.h
```diff
@@ -3050,35 +3053,69 @@ namespace oomph
...
+    {
+      return Linear_solver_error;
+    }
+
+    /// Access function to Max. # of iterations performed when the Newton solver
+    /// died
+    unsigned get_iterations()
```
MatthiasHeilManchester, line 3110: same: kill "get_"

src/generic/problem.h
```diff
@@ -3050,35 +3053,69 @@ namespace oomph
...
+    {
+      return Iterations;
+    }
+
+    /// Access function to Max. residual when Newton solver died
+    double get_maxres()
```
MatthiasHeilManchester, line 3116: and another unnecessary "get_"

PR #121, opened 2024-04-16
Varying the minimum permitted angle in the triangle mesh generation.

PaulAlexander23:
Add an internal variable to store the minimum permitted triangle angle.
Add a setter and a getter.
Use the getter within the class where it is needed.
Note: I have used magic numbers in the constructor as I don't know the best place to locate a default variable in the templated class.

### DIFF
--- a/src/meshes/triangle_mesh.template.h
+++ b/src/meshes/triangle_mesh.template.h
@@ -424,7 +424,7 @@ namespace oomph
   {
   public:
     /// Empty constructor
-    TriangleMesh()
+    TriangleMesh() : Min_element_interior_angle(30.0)
     {
 #ifdef OOMPH_HAS_TRIANGLE_LIB
       // Using this constructor no Triangulateio object is built
@@ -451,6 +451,7 @@ namespace oomph
       const std::string& poly_file_name,
       TimeStepper* time_stepper_pt = &Mesh::Default_TimeStepper,
       const bool& allow_automatic_creation_of_vertices_on_boundaries = true)
+      : Min_element_interior_angle(30.0)
     {
       // Mesh can only be built with 2D Telements.
       MeshChecker::assert_geometric_element<TElementGeometricBase, ELEMENT>(2);
@@ -515,6 +516,7 @@ namespace oomph
     /// TriangleMeshParameters
     TriangleMesh(TriangleMeshParameters& triangle_mesh_parameters,
                  TimeStepper* time_stepper_pt = &Mesh::Default_TimeStepper)
+      : Min_element_interior_angle(30.0)
     {
       // Store the region target areas
       Regions_areas = triangle_mesh_parameters.target_area_for_region();
@@ -757,6 +759,7 @@ namespace oomph
       const double& element_area,
       TimeStepper* time_stepper_pt = &Mesh::Default_TimeStepper,
       const bool& allow_automatic_creation_of_vertices_on_boundaries = true)
+      : Min_element_interior_angle(30.0)
     {
       // Mesh can only be built with 2D Telements.
       MeshChecker::assert_geometric_element<TElementGeometricBase, ELEMENT>(2);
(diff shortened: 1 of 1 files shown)


PaulAlexander23:
Comment made in oomph-lunch.
Move this to the triangle parameters class.

PaulAlexander23:
- All calls to the triangle API require the minimum permitted element angle to be specified.
- The triangle API is called in the TriangleMesh constructor
- The triangle parameters class seems appropriate, but not all the Triangle meshes are constructed using the triangle parameter class.
- The TriangleIO struct is passed into the triangle API separately from the minimum element angle, so this parameter can't be stored inside that object either.
Therefore I have left the minimum permitted triangle angle in the TriangleMesh class.

MatthiasHeilManchester:
Hmmm; I've now looked at the code and the resulting behaviour is a bit odd: Given that before the mesh is built (by calling its constructor) there's no way to access the member function that sets the mimimum angle, the mesh is always built with the default minimum angle. Changing that after the mesh has been constructed has zero effect -- unless it's a refineable mesh in which case the changed minimum angle is enforced (only!) when we're adapting (maybe; haven't thought this through completely). if this is the intended/actual behaviour the function that sets the minimum angle should be called something like set_min_angle_for_next_adapation(...) (bit of a mouthful, but that's what it does) and it that should only live in the refineable version. 

All a bit odd. Wouldn't it be much cleaner to add the minimum angle to the Parameter class and add comments to the constructors that don't use it that the mesh is built with a default minimum angle (or not say it at all since this is understood and is no change in behaviour). I may want to set minimum angles even for non-refineable meshes.

PaulAlexander23:
Yep. I agree. The previous code was a bit half-baked.
I've updated the code now so you can pass in the min element angle to the triangle mesh parameters for the initial mesh generation.
I've also added a function to the refineable triangle mesh to allow setting the min element angle for the subsequent adaptions.

src/meshes/triangle_mesh.template.cc
```diff
@@ -30260,6 +30260,10 @@ namespace oomph
       // Set the element area
       triangle_mesh_parameters.element_area() = max_area;
 
+      // Set the minimum element interior angle.
+      triangle_mesh_parameters.set_min_element_interior_angle(
+        this->Min_element_interior_angle_for_adaption);
```
alhazel, line 30265: Min_element_interior_angle_after_adaptation is (I think) more accurate.

src/meshes/triangle_mesh.template.h
```diff
@@ -2491,6 +2528,18 @@ namespace oomph
...
     }
 
+    /// Read-only access to the min element interior angle for adaption
+    const double& min_element_interior_angle_for_adaption() const
+    {
+      return this->Min_element_interior_angle_for_adaption;
+    }
+
+    /// Set the min element interior angle for adaption
+    void set_min_element_interior_angle_for_adaption(const double& new_angle)
```
alhazel, line 2538: Definitely think that using "after" rather than "for" conveys the sense better for this function. Presumably this can be used at any point between adaptations as a means to change the minimum angle.

MatthiasHeilManchester: TBH, I still don't really understand why we can't just include this parameter into the arguments class that passes all the other parameters (with sensible defaults for those that aren't explicitly specified) to the constructor. Distinguishing between what happens when the mesh is first generated and during the next adaptation seems a bit odd (and implies that it's a hacky work-around for not being able to set the parameter straightaway. Why would I want to do the first calculation on a mesh that I don't like long term?

src/meshes/triangle_mesh.template.cc
PaulAlexander23: The angle is required for the adaption step. Perhaps Min_element_interior_angle_during_adaption is better, or to just drop the adaption part: Min_element_interior_angle.

PaulAlexander23: Min_permitted_element_interior_angle would be better still as it is a tolerance rather than a measurement.

src/meshes/triangle_mesh.template.h
PaulAlexander23: I am not distinguishing between first generated mesh and the subsequent adaption steps. If you want to, you can use the set_min_element_interior_angle(new_angle).

There are three constructors for Triangle meshes: a default one with no arguments, one with our lovely arguments class `TriangleMeshParameters` and one specifying a node file name, element file name and poly file name.
The parameters class is not stored within `TriangleMesh` so the min interior angle would be lost after constructor unless we store it within the class.
We need to pass it as an optional argument in the other cases as the mesh gets built in the constructor, so we need this parameter in the constructor.

PaulAlexander23:
I still need to add demo drivers to demonstrate this working.

PR #122, opened 2024-04-23
Eigenvalue cleanup and make repeated eqn number a warning.

MatthiasHeilManchester:
- Finally got rid of ARPACK and associated legacy code  everywhere. 

- Made error that is thrown when a repeated equation number is detected in an element a warning rather than error (there are reasons when this is sane);  warning can be made to shut up by setting GeneralisedElement::Suppress_warning_about_any_repeated_data to true.

- Added proper self-test for Rupinder's faceted surface vtu style output (ignoring anything that involves triangle because results then become non-deterministic).

- Ubuntu passes cleanly: 

      https://github.com/MatthiasHeilManchester/oomph-lib/actions/runs/8800453758

- Mac fails on the elastic Bretherton code (long standing mysterious problem that only occurs sometimes!): 

       https://github.com/MatthiasHeilManchester/oomph-lib/actions/runs/8800453762


self_test/faceted_surface_vtu_output/faceted_surface_vtu_output_tester.cc
```diff
@@ -93,16 +98,22 @@ int main()
...
+
+//   // DiskWithTwoLayersTetMeshFacetedSurface
+//   //----------------------------------------
+//   // Create a disk with two layers, looks like a coin
+//   DiskWithTwoLayersTetMeshFacetedSurface disk_with_two_layers_faceted_surface(
+//     &disk, half_nsegment, dummy_id, dummy_id, dummy_id, dummy_id, dummy_id);
+
+//   // Output into the Paraview .vtu format
+//   disk_with_two_layers_faceted_surface.output_paraview(
+//     "disk_with_two_layers_faceted.vtu");
```
PaulAlexander23, line 116: Could we remove the commented-out code? Or add a flag for non-deterministic tests? Or move it to its own script.

self_test/faceted_surface_vtu_output/faceted_surface_vtu_output_tester.cc
```diff
@@ -54,27 +54,32 @@ int main()
...
-  DiskTetMeshFacetedSurface disk_faceted_surface(
-    &disk, half_nsegment, dummy_id, dummy_id);
-  // Output into the Paraview .vtu format
-  disk_faceted_surface.output_paraview("disk_faceted.vtu");
+  // // Create the faceted surface
+  // DiskTetMeshFacetedSurface disk_faceted_surface(
+  //   &disk, half_nsegment, dummy_id, dummy_id);
+  
+  // // Output into the Paraview .vtu format
+  // disk_faceted_surface.output_paraview("disk_faceted.vtu");
```
PaulAlexander23, line 82: See comment on commented-out code.

config/configure_options/private_configure_options/xeon_biolaptop_mpi
```diff
@@ -124,15 +124,15 @@ LD=mpif77
...
-#--with-hypre=/home/mheil/local/hypre_default_installation_mpi
-#--with-trilinos=/home/mheil/local/trilinos_default_installation_mpi
+#--with-blacs=/home/matthias/local/mumps_and_scalapack_default_installation/lib
+#--with-scalapack=/home/matthias/local/mumps_and_scalapack_default_installation/lib/libscalapack.a
+#--with-pord=/home/matthias/local/mumps_and_scalapack_default_installation/lib/libpord.a 
+#--with-mumps=/home/matthias/local/mumps_and_scalapack_default_installation
+#--with-blas=/home/matthias/local/lib/blas/blas.a
+#--with-lapack=/home/matthias/local/lib/lapack/lapack.a
+#--with-hypre=/home/matthias/local/hypre_default_installation_mpi
+#--with-trilinos=/home/matthias/local/trilinos_default_installation_mpi
```
PaulAlexander23, line 134: Is there a better way of handling the configure options rather than adding private configure options to the public repo?

PaulAlexander23 (approved):
All the changes make sense to me.
I haven't gone through the rest of the codebase looking for missed changes. Let me know if this is something you would like me to do.
I've given a couple of minor comments too.

self_test/faceted_surface_vtu_output/faceted_surface_vtu_output_tester.cc
MatthiasHeilManchester: Done.

self_test/faceted_surface_vtu_output/faceted_surface_vtu_output_tester.cc
MatthiasHeilManchester: done.

demo_drivers/bifurcation_tracking/adaptive_hopf_with_separate_meshes.cc
```diff
@@ -60,12 +60,11 @@ namespace Global_Parameters
...
 
- // Control Flag that will read in the eigenfunction from disk
- // to avoid a length eigenvalue calculation while testing
- bool Read_in_eigenfunction_from_disk = true;
-
  //Set the normalisation for the eigenfunction
  std::complex<double> Eigenfunction_normalisation(1.0,0.0);
+ 
+ // Control Flag that will read in the eigenfunction from disk
+ bool Read_in_eigenfunction_from_disk = true;
```
alhazel, line 67: Comment is a bit loose. The control flag doesn't read in the eigenfuction. Suggest.

"// Control flag used to determine whether the eigenfunction is read from disk. 
 // Reading from disk avoids lengthy eigenvalue calculations in the self-tests"

alhazel:
This is likely to break a number of codes that used the legacy interfaces for stability calculations, so can it be labelled in any way to make that clear to people? Ultimately, those codes should be updated, or check out older versions of the library.

MatthiasHeilManchester:
> This is likely to break a number of codes that used the legacy interfaces for stability calculations, so can it be labelled in any way to make that clear to people? Ultimately, those codes should be updated, or check out older versions of the library.

Well, yes, it's an interface change, but the old interface never really made any sense so I'm not sure there's a reason for keeping it alive. If you're worried about backwards compatibility, then we should possibly do the appropriate version number bump to indicate a change to an interface. Not sure how religious we've been about this so far. I suspect not always.

demo_drivers/bifurcation_tracking/adaptive_hopf_with_separate_meshes.cc
MatthiasHeilManchester: changed, commited and pushed.

PR #123, opened 2024-05-03
Minor fixes for the linear stability elements

PaulAlexander23:
Force compilation of 1D elements. Needed for one of my projects but perhaps doesn't need to be included here.
Extract implementation to a cc file, as I was getting a "multiple definitions" error of this function.


alhazel:
Seems uncontroversial to me.

PR #129, opened 2024-10-18
Fix the clang format workflow

PaulAlexander23:
Using the latest version of ubuntu causes and issue with the clang format workflow.
I've fixed the issue and completed a pull request such that the latest of Puneet's workflows includes the fix.
If the tests pass then the fix has worked.
There might also be an issue with Python2 (needed for installing trilinos) for the latest Ubuntu version 24.04.1.


PaulAlexander23:
Converted to a draft until the tests pass.

PaulAlexander23:
Added tests to run on Ubuntu 22.04.

.github/workflows/clang-format.yaml
```diff
@@ -22,7 +25,7 @@ jobs:
 
       # Format all files in the src/ directory except cfortran.h, lapack_qz.h,
       # superlu.c, superlu_dist.c, which sit in the src/generic/ subdirectory.
-      - uses: PuneetMatharu/clang-format-lint-action@v0.13
+      - uses: PuneetMatharu/clang-format-lint-action@master
```
puneetmatharu, line 28: Use `v0.18`, not the branch name.

.github/workflows/self-tests-ubuntu-22-04.yaml
```diff
@@ -0,0 +1,91 @@
+name: Ubuntu 22.04 self-tests
```
puneetmatharu, line 1: This is _not_ the way to run self-tests on multiple versions of Ubuntu. Creating a new workflow file creates a large amount of code duplication. We also don't want things to diverge between the different Ubuntu-target self-test workflows, and thus have to manage twice as many workflows. Look at using a [matrix](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#using-a-matrix-strategy).

README.md
```diff
@@ -23,7 +23,7 @@
         <th><a href="../../tree/development"><code>development</code></a></th>
     </tr>
     <tr>
-        <td>Ubuntu</td>
+        <td>Ubuntu Latest (24.04)</td>
```
puneetmatharu, line 26: Please undo the changes in the README. We can just test for multiple OS versions within the same workflow and group it under Linux.

puneetmatharu (changes requested):
Thanks for creating a PR to handle this, but I think we can use a much simpler approach IF we want to truly test on multiple OS versions (which includes burning more fossil fuels to run the self-tests). @MatthiasHeilManchester do you have any opinions on this?

puneetmatharu:
Great. Thank you!

MatthiasHeilManchester:
Just catching up with this... Agree that we should try to keep things going with the latest ubuntu (though I've heard bad things about it and was told not to upgrade...). 

Googled around a bit and found a long discussion here:

https://github.com/actions/setup-python/issues/672

which suggests this:

runs-on: ubuntu-20.04
    container:
      image: python:2.7.18-buster

Don't understand it well enough to be able to comment...

   m.

PR #131, opened 2024-10-24
Patch broken macOS build and missing python2.7 in GitHub actions

puneetmatharu:
### Description

Fixes the issue with Python 2.7 being missing from GitHub Actions; a temporary solution for now. We should move to a more recent version of Trilinos to handle the missing Python 2.7 which has reached End of Life on most OSs. I've also handled an error arising with the macOS build, where C99 errors were springing up. Have added a new macOS config (for the GitHub Actions only) that adds the necessary `-std=c90` C compiler flag in. This will become redundant once we switch to the CMake build.

### Self-tests

Ubuntu tests pass with and without MPI but [macOS fails](https://github.com/puneetmatharu/oomph-lib/actions/runs/11486946942) on 8-9 tests. However, macOS can be patched at a later time; they're not the highest priority right now and go outside the scope of this PR.

- [x] [Ubuntu self-tests](https://github.com/puneetmatharu/oomph-lib/actions/workflows/self-tests-ubuntu.yaml)
- [x] [macOS self-tests](https://github.com/puneetmatharu/oomph-lib/actions/workflows/self-tests-macos.yaml)

*_checkbox indicates whether the latest tests on the branch have completed._


puneetmatharu:
@MatthiasHeilManchester happy to look at this PR? Should be a quick one. (I'm half tempted to just merge it in myself...)

PaulAlexander23:
Looks good to me (for what that's worth). Thanks, Puneet

PR #133, opened 2024-11-05
Trigger the self tests to run once every month

PaulAlexander23:
In case any of the dependencies are updated in the meantime, such as the ubuntu-latest.
This means that the homepage readme will display up to date information (well within the last month) about the main branch of the repository.

The schedule run (similar to the update copyright run) only runs on the default (main) branch. So if people fork and create there own branches then only the main branches will run each month, to give an idea of the environmental cost. We should therefore merge the latest development branch into the main branch after this pull request, unless there is any good reason not too. (Or change this pull request to pull into main!) We should probably update the minor version in the tags too.

Perhaps there is a better way to implement this, and I am open to suggestions and happy to implement changes.

### DIFF
--- a/.github/workflows/self-tests-macos.yaml
+++ b/.github/workflows/self-tests-macos.yaml
@@ -1,6 +1,13 @@
 name: macOS self-tests
 
-on: push
+# IMPORTANT: A schedule workflow action *must* be placed on the default branch
+# of the repository. Only then will it run as scheduled. It is also placed in a
+# queue when scheduled, so it may not run exactly at the requested time.
+on:
+  push: # Run when the branch is pushed to
+  schedule:
+    - cron: '0 0 1 * *'  # Every month on the first at midnight UTC
+  # Calculated using crontab.guru. See here: https://crontab.guru/#1_0_1_1_*.
 
 jobs:
   build:
--- a/.github/workflows/self-tests-ubuntu.yaml
+++ b/.github/workflows/self-tests-ubuntu.yaml
@@ -1,7 +1,13 @@
 name: Ubuntu self-tests
 
-on: push
-
+# IMPORTANT: A schedule workflow action *must* be placed on the default branch
+# of the repository. Only then will it run as scheduled. It is also placed in a
+# queue when scheduled, so it may not run exactly at the requested time.
+on:
+  push: # Run when the branch is pushed to
+  schedule:
+    - cron: '0 0 1 * *'  # Every month on the first at midnight UTC
+  # Calculated using crontab.guru. See here: https://crontab.guru/#1_0_1_1_*.
 jobs:
   build:
     strategy:


PaulAlexander23:
- [ ] Run problematic tests on macbook and update

PaulAlexander23:
- [ ] Send tar file from linearised_porous.. to Matthias

PaulAlexander23:
> * [x]  Run problematic tests on macbook and update

I've run the problematic test on my macbook and it passed just fine.

MatthiasHeilManchester:
> The tests pass!

Yay! But still waiting for the tar file for the larger-than-ideal change to some of the tolerances (elastic Bretherton I think).

PaulAlexander23:
Uh oh. My macbook tests pass, but now the Github ones fail. It looks like a github workflow Heisen-bug, and I can't replicate it on my macbook.
Any suggestions?

MatthiasHeilManchester:
> Uh oh. My macbook tests pass, but now the Github ones fail. It looks like a github workflow Heisen-bug, and I can't replicate it on my macbook. Any suggestions?

Can we have a chat tomorrow (Fri) I'm in the office most of the day (teaching from 12-1).

  M.

PaulAlexander23:
Update after in-person meeting:
- [ ] Matthias to look at the results tar file and ensure that the macOS results are correct.
- [ ] Paul to look at how to run the elastic Bretherton for macOS only, or to not include this test.

PaulAlexander23:
Where would be the best place to put the error logs generated from the test when it fails?

demo_drivers/interaction/elastic_bretherton/validate.sh
```diff
@@ -17,33 +17,45 @@ mkdir Validation
...
-echo " " >> validation.log
-echo "Validation directory: " >> validation.log
-echo " " >> validation.log
-echo "  " `pwd` >> validation.log
-echo " " >> validation.log
-cat  RESLT/trace.dat  RESLT/soln2.dat > el_breth.dat
-
-if test "$1" = "no_fpdiff"; then
-  echo "dummy [OK] -- Can't run fpdiff.py because we don't have python or validata" >> validation.log
+if [[ "$(uname)" == "Darwin" ]]; then
```
PaulAlexander23, line 20: Only run this test on macOS

bin/change_headers_to_links.sh
```diff
@@ -2,7 +2,7 @@
 
 #---------------------------------------------------------------------------
 # Little helper script to turn include header files into symbolic
-# links -- needed because we need to prefix the sources for ln -s [in 
+# links -- needed because we need to prefix the sources for ln -sf [in 
 # its $(LN_S) incarnation] with their full directory.
```
PaulAlexander23, line 6: Including `-f` for the soft linking command `ln -s` ensures that the link is created even if we are overwriting the file. This was causing rebuilding the library to fail when testing on my macbook.
Let me know if you want me to remove these forcing flags.

demo_drivers/interaction/linearised_poroelastic_fsi_pulsewave/validate.sh
```diff
@@ -44,7 +44,7 @@ if test "$1" = "no_fpdiff"; then
   echo "dummy [OK] -- Can't run fpdiff.py because we don't have python or validata" >> validation.log
 else
 ../../../../bin/fpdiff.py ../validata/results.dat.gz   \
-         results.dat  0.5 1.0e-10 >> validation.log
+         results.dat  0.5 3.0e-4 >> validation.log
```
PaulAlexander23, line 47: This is the large difference

PR #141, opened 2025-02-10
Tetmesh regions and boundary-lookup functionality

wobblygomboc:
add helper functions regenereate_region_boundary_lookups() and add_element_in_region_pt to manually force a full regeneration of region boundary lookup schemes and to manually add an element to a Tetmesh's region


MatthiasHeilManchester (changes requested):
Odd: the changes to  src/meshes/refineable_tetgen_mesh.template.cc appear over and over again. Have already pulled them in twice (without problem, as far as I can tell). ditto with the QElement<3,5> which features again here.

 Anyway, this one has a few real queries/change requests.

wobblygomboc:
still don't seem to have fully mastered this git workflow, files from other commits seem to keep appearing in the wrong branch commits!

PR #143, opened 2025-02-10
Mumps updates

wobblygomboc:
add parallel ordering options (switch on ParMetis and PT_Scotch)
improve error identification - currently all errors are assumed to be workspace errors, which causes an infinite loop if the jacobian is singular. Numerically and structurally singular matrices are now identified separately, with the MUMPS magic numbers stored in a private enum in the MumpsSolver class.


wobblygomboc:
I've re-commited tet_mesh.[h/cc], realised my ineptitude with git was about to cause issues if you merge the tetmesh PR first because these files would get changed back by this PR.

src/generic/mumps_solver.h
```diff
@@ -238,6 +238,18 @@ namespace oomph
       Jacobian_ordering_flag = Metis_ordering;
     }
 
+    // tell MUMPS to use the METIS package for the ordering
```
MatthiasHeilManchester, line 241: All good stuff; is  pscotch always available or does that require the provision of yet another third party library?

wobblygomboc: yeah it will require the user to have ptscotch separately on their machine and have linked against it appropriately during their oomph-lib build...just seemed useful to provide an oomph-lib interface for switching on the options that mumps offers. It also isn't risky / assuming the user has set everything up perfectly - mumps is clever internally, these switches act like a request rather than a demand, so if you ask for an ordering package that it can't find it will just default back to whatever it has rather than dying.

PR #166, opened 2025-08-07
Cmake beta

MatthiasHeilManchester:
A few clarifications, mainly re how to link stand-alone codes against oomph-lib. These need checking by @puneetmatharu . Also started to update the list of publications and they're the main reason I'm sending this PR now. Andrew will add more and there's a good chance of avoidable clashes if we're not careful, so I'm trying to push this now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,7 @@
     - [Autotools](#autotools-5)
     - [CMake](#cmake-6)
 - [Dos and Don'ts](#dos-and-donts)
+- [FAQs](#faq)
 - [Additional information for developers](#additional-information-for-developers)
   - [Use symbolic links for header files](#use-symbolic-links-for-header-files)
   - [Creating robust `validata` for self tests](#creating-robust-validata-for-self-tests)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1012,6 +1017,13 @@ oomph_add_executable(
 # ------------------------------------------------------------------------------
 ```
 
+> [!NOTE]
+> In the above example we have included the header file `mesh_stuff.h` into the `SOURCES`
```
puneetmatharu, line 1021: Do you mean `glued_mesh_stuff.h`?

README.md
```diff
@@ -1047,10 +1065,44 @@ mkdir RESLT
 ./mesh_gluing
 ```
 
+Note the three options for specifying the install directory (but stick to Option 1!).
+
+Given that the install directory is quite deep, users sometimes worry if they 
+have specified the right level of that directory tree. If you've only just installed `oomph-lib` 
+it's easy to reconstruct:
+- If you've built `oomph-lib` by running `oomph_build.py` (or the corresponding `cmake` 
+  commands) in `/home/joe_cool/oomph-lib`, say, without explicitly  specifying an install
```
puneetmatharu, line 1074: A bit *ahem* particular, but there's a double space after "explicitly"

README.md
```diff
@@ -1047,10 +1065,44 @@ mkdir RESLT
...
+  ``` 
+  then this is the one you use.
+
+If you're unsure (because the installation was too long ago, say) you should check 
+that the directory you're about to specify as the install directory contains the 
+file `oomphlibConfig.cmake`. This is the file that contains the key information
+that allows CMake to work with the installed library. So doing this
+```bash
+# Check that oomphlibConfig.cmake is there
+find  /home/joe_cool/local/oomph-lib -name 'oomphlibConfig.cmake'
```
puneetmatharu, line 1093: Double space after `find` (sorry!)

README.md
```diff
@@ -1047,10 +1065,44 @@ mkdir RESLT
...
+```
+should find the file
+```bash
+/home/joe_cool/local/oomph-lib/lib/cmake/oomphlib/oomphlibConfig.cmake
+```
+if the file can't be found (or it appears at a different level in the directory 
+tree the configuration will fail.
```
puneetmatharu, line 1100: closing bracket after "tree"

puneetmatharu (changes requested):
Looks great! I can only propose some very minor tweaks.

README.md
MatthiasHeilManchester: Gone!

README.md
MatthiasHeilManchester: Also gone!

MatthiasHeilManchester:
Thanks for all those spots. No such thing as minor issues! Have pushed the changes.

MatthiasHeilManchester:
Merged in myself (naughty!) to make sure Andrew's got the latest source. I did deal with everything; honest!

Issue #149, opened 2025-06-17
v3.0 (cmake) beta tester issues


MatthiasHeilManchester:
Add your comments below, please. We'll tick them off when they've been dealt with.

puneetmatharu:
> - [ ] Add list of third party libraries to README.md (copy from copyright.txt).

Done. See [here](https://github.com/oomph-lib/oomph-lib/compare/feature-shared-cmake...puneetmatharu:oomph-lib:feature-shared-cmake). Note that my MUMPS (which is used by oomph-lib) is borrowed from [`scivision/mumps`](https://github.com/scivision/mumps), which is just clones the version of MUMPS from [the main webpage](https://mumps-solver.org/index.php) and wraps it in CMake machinery.

I needed to create my own version so that it could find LAPACK properly. It's a little messy. I think I can potentially use the scivision repo directly but it's going to be painful and I don't want to invest the time right now.

MatthiasHeilManchester:
Delete:
-  non_interactive_autogen.sh
-  validate.sh

in top level oomh-lib directory

MatthiasHeilManchester:
> N.B. I think you should be able to use
> 
> cmake [...] -DCMAKE_CXX_FLAGS="-DOOMPH_3_5_BRICK_FOR_MESHING_ONLY_NO_INTEGRATION"
> or
> 
> ./oomph_build.py [...] --oomph-OOMPH_INSTALL_HEADERS_AS_SYMLINKS=ON \
>                  --oomph-extra-flags="-DCMAKE_CXX_FLAGS=\"-DOOMPH_3_5_BRICK_FOR_MESHING_ONLY_NO_INTEGRATION\""
> Noting that you're passing the extra CMake option inside a string to `--oomph-extra-flags`. I haven't tested it this morning though. It might also require a semicolon separated list for multiple options. I've added a clean option in [#167](https://github.com/oomph-lib/oomph-lib/pull/167) for handling a space separated list.

Should we remove this option now? I'd spotted it but didn't quite understand the instructions. If it achieves the same as the newly added option let's kill this one. They key question is always if this flag is passed through to stand-alone driver codes or not.

MatthiasHeilManchester:
> Delete:
> 
>     * non_interactive_autogen.sh
> 
>     * validate.sh
> 
> 
> in top level oomph-lib directory

Done by @puneetmatharu "Closing" ( = hiding) this now.

puneetmatharu:
[...]

PR #168, opened 2025-08-13
Rename `--oomph-extra-flags` -> `--oomph-extra-cmake-options` in `oomph_build.py`

puneetmatharu:
To eliminate the confusion in https://github.com/oomph-lib/oomph-lib/issues/149#issuecomment-3179292193. This flag is meant to be used to pass CMake options, not C/C++ specific flags. I've renamed the command line flag for `oomph_build.py` accordingly.

### DIFF
--- a/oomph_build.py
+++ b/oomph_build.py
@@ -2,7 +2,7 @@
 
 import os
 import json
-import re
+import shlex
 import shutil
 import subprocess
 import sys
@@ -68,7 +68,8 @@ def run_command(cmd: list[str], cwd: Path, verbose: bool):
     try:
         if not verbose:
             result = subprocess.run(
-                cmd, cwd=cwd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, text=True)
+                cmd, cwd=cwd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, text=True,
+            )
         else:
             result = subprocess.run(cmd, cwd=cwd)
     except Exception as e:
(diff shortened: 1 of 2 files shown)


oomph_build.py
```diff
@@ -488,7 +488,7 @@ def expanded_path(p):
     ext_group.add_argument("--ext-OOMPH_USE_BOOST_FROM", type=expanded_path, metavar="PATH", help="Use a preinstalled Boost from the given path; used by CGAL.")
 
     # Any additional flags for the external distributions project
-    ext_group.add_argument("--ext-extra-flags", nargs="+", metavar="FLAG", help="Additional raw CMake flags for external_distributions (e.g. -DXYZ=VALUE).")
+    ext_group.add_argument("--ext-extra-cmake-options", nargs="+", metavar="FLAG", help="Additional raw CMake flags for external_distributions (e.g. -DXYZ=VALUE).")
```
MatthiasHeilManchester, line 491: I always struggle with these overly generic placeholders. Can we specify a real example of an option one may actually want to pass to cmake; ideally something that users are likely to have come across already and will immediately recognise as something that would only be passed to cmake.

oomph_build.py
```diff
@@ -506,7 +506,7 @@ def expanded_path(p):
     oomph_group.add_argument("--oomph-OOMPH_EXTRA_COMPILE_DEFINES", metavar="DEFINES", help="Additional C/C++ compile defines to pass to the oomph-lib build.")
 
     # Any additional flags for oomph-lib
-    oomph_group.add_argument("--oomph-extra-flags", nargs="+", metavar="FLAG", help="Additional raw CMake flags for oomph-lib (e.g. -DXYZ=VALUE).")
+    oomph_group.add_argument("--oomph-extra-cmake-options", nargs="+", metavar="FLAG", help="Additional raw CMake flags for oomph-lib (e.g. -DXYZ=VALUE).")
```
MatthiasHeilManchester, line 509: DUPLICATED COMMENT: I always struggle with these overly generic placeholders. Can we specify a real example of an option one may actually want to pass to cmake; ideally something that users are likely to have come across already and will immediately recognise as something that would only be passed to cmake.

PR #179, opened 2025-10-17
README.md in external distributions (this time the real one!)

MatthiasHeilManchester:


external_distributions/README.md
```diff
@@ -63,21 +63,23 @@ If you cannot obtain a recent enough version of CMake via your favourite package
...
-| `SuperLU_DIST`                   | [v9.1.0](https://github.com/xiaoyeli/superlu_dist/tree/v9.1.0)                                                                            |
-| `MUMPS`                          | [5.6.2](https://github.com/puneetmatharu/mumps/tree/v5.6.2.5)                                                                             |
-| `HYPRE`                          | [2.32.0](https://github.com/hypre-space/hypre/tree/v2.32.0)                                                                               |
-| `Trilinos`                       | [16.0.0](https://github.com/trilinos/Trilinos/tree/trilinos-release-16-0-0)                                                               |
+The table below lists the version of the third party libraries that work with(and/or are required by) `oomph-lib` and indicates which ones we can install for you via CMake. By default, we build "everything we can"; see below for how to customise the build process and select a serial or an MPI build. If you wish to provide your own version of these libraries, make sure it is the right version.
```
puneetmatharu: > lists the version of the third party libraries that work with(and/or are required by) `oomph-lib` and indicates which ones we can install for you via CMake

Hmm, is it just me or does this give the impression that we can only install a subset of the libraries that `oomph-lib` works with? Even though we can install all the ones we typically work with?

Would something like the following be clearer?

> `oomph-lib` depends on and works with a number of third-party libraries however we can install these libraries for you with CMake. In the table below we list these libraries and the versions that we can install for you.

puneetmatharu: > see below for how to customise

Hmm, I'm guessing you're not referring to the table here. Maybe "see further below" might be clearer?

external_distributions/README.md
```diff
@@ -63,21 +63,23 @@ If you cannot obtain a recent enough version of CMake via your favourite package
...
+
+
+| Library        | Required/optional | Built by default (serial build)? | Built by default (MPI build)?  | Version |
+| ----           | ---               | -----                    | ---                    | ---     | 
+| `OpenBLAS`     | required by `oomph-lib`              | Yes (though not on macOS; see below) | Yes (though not on macOS; see below) |  [0.3.25](https://github.com/OpenMathLib/OpenBLAS/tree/v0.3.29)      |
+| `SuperLU`       | required by `oomph-lib`               | Yes | Yes | [v6.0.1](https://github.com/xiaoyeli/superlu/tree/v6.0.1) | 
+| `METIS`        | required by `oomph-lib` (via `SuperLU`) | Yes | Yes | [commit `a6e6a2cfa92f93a3ee2971ebc9ddfc3b0b581ab2`](https://github.com/KarypisLab/METIS/tree/a6e6a2cfa92f93a3ee2971ebc9ddfc3b0b581ab2)  |
```
puneetmatharu: Admittedly these commit hashes are a little obtrusive... We can shorten these to the first 7 characters. Would you be able to update the rows that have commit hashes in them?

external_distributions/README.md
puneetmatharu: Can we update this to say

```bash
>>> ls -1 /opt/homebrew/opt/boost
```

Technically it's still correct but stripping the `@1.87` will make the later command

```bash
... -DOOMPH_USE_BOOST_FROM=$(brew --prefix boost) ...
```

clearer

puneetmatharu (changes requested):
Huge thank you for taking care of these! Just a few trivial (but hopefully useful) remarks from me!

external_distributions/README.md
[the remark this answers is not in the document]
MatthiasHeilManchester: Done.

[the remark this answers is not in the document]
MatthiasHeilManchester: I'd wondered about that while searching for "homebrew". Done, and removed the trailing "/" too. Not sure if it matters.

puneetmatharu:
If I remember correctly, hlib didn't even do anything but GMP/MPFR were thrown away to use the Boost Multiprecision library (see comment [here](https://github.com/oomph-lib/oomph-lib/blob/67d296d6f326ac7b725f83a11895888f0a7d33b0/external_distributions/cmake/OomphGetExternalCGAL.cmake#L58))

PR #181, opened 2025-10-21
Cmake beta --> development

MatthiasHeilManchester:
Over to Andrew: 4+ years in the making (with some dormant periods along the way!). Puneet and I are happy for this to become the new main,

### DIFF
--- a/.github/workflows/self-tests-macos.yaml
+++ b/.github/workflows/self-tests-macos.yaml
@@ -116,8 +116,8 @@ jobs:
         with:
           path: |
             external_distributions/install
-            external_distributions/build/cmake_flags_for_oomph_lib.txt
-            external_distributions/build/cmake_flags_for_oomph_lib.json
+            external_distributions/install/cmake_flags_for_oomph_lib.txt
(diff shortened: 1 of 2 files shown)


puneetmatharu:
Looks like the running out of disk space error is a genuine issue. I think I might have a solution for it. Testing it now...

MatthiasHeilManchester:
> Looks like the running out of disk space error is a genuine issue. I think I might have a solution for it. Testing it now...

@puneetmatharu : Ooops; that's a new one! (Just had a look to check that the tests had a all passed and the big green button can be pressed; not quite yet!). Fingers crossed...

MatthiasHeilManchester:
OK; thanks that's really useful and in that case, I think we're ready to go. It's only the push to development branch anyway :) About to press, the button -- watch out for that mushroom-shaped cloud appearing on the horizon.

puneetmatharu:
> OK; thanks that's really useful and in that case, I think we're ready to go. It's only the push to development branch anyway :) 

Yay!

> About to press, the button -- watch out for that mushroom-shaped cloud appearing on the horizon. 

Just got to push to 'main' and we're done! Only took a few years...

PR #188, opened 2025-11-27
Make GitHub actions run on PRs

puneetmatharu:


puneetmatharu:
Notice how we get self-tests running in the PRs itself rather than relying on the user posting links to self-tests. The concurrency also places a limit so that we can only run a workflow once per branch/PR.

Issue #191, opened 2025-11-28
Should oomph_build.py have a (command line) option to specify the build directory?


MatthiasHeilManchester:
Should oomph_build.py have a (command line) option to specify the build directory? I don't think it does; it always goes to build in the oomph-lib home directory. The only option I could see is

    ./oomph_build.py [...]  --oomph-extra-cmake-options=' -B build_debug_paranoid_junk_test'
 
but this obviously doesn't work, and I can't see any other flags that would give me a hook into this.

puneetmatharu:
Handled by https://github.com/oomph-lib/oomph-lib/pull/192.

PR #192, opened 2025-12-01
Let user override the build dirs in `oomph_build.py`

puneetmatharu:
Added the options requested in https://github.com/oomph-lib/oomph-lib/issues/191. Admittedly they're not currently tested with the CI; we just use the defaults, so I'm not sure we can use them to check everything still works. It seemed to build fine with my updates locally though.


MatthiasHeilManchester (approved):
Thanks for doing this!

MatthiasHeilManchester:
Checked it out and it works like a dream!

PR #196, opened 2025-12-06
Update MUMPS support.

puneetmatharu:
### Summary of changes

 - Uses latest updates from my MUMPS fork (to pull in the latest changes from the main mumps repo).
 - Switched to the latest upstream MUMPS version (5.8.1).
 - Add workflows to cache MUMPS tarball and use across jobs. By caching it we reduce the number of times we need to download the tarball so we don't hit the download limit when running PRs.


puneetmatharu:
Merging now because I've already tested it before creating the PR and I want to rebase my other PRs on this change.

PR #198, opened 2025-12-06
Update MUMPS support

puneetmatharu:
Updates include:
 - Updated to use the latest changes from my MUMPS fork (pulls in the latest changes from the main MUMPS repo which means we can provide our own MUMPS tarball).
 - Switched to latest upstream MUMPS version (i.e. the tarball version).
 - Add composite actions to cache MUMPS tarball (`.github/actions/cache-mumps/action.yaml`) and reuse (`.github/actions/restore-mumps/action.yaml`) across jobs.

### DIFF
--- a/.github/actions/cache-mumps/action.yaml
+++ b/.github/actions/cache-mumps/action.yaml
@@ -1,0 +1,48 @@
+--- # ------------------------------------------------------------------------------
+
+name: Cache MUMPS tarball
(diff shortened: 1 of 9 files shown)


puneetmatharu:
Tests pass and this time there's no dependence on my own repo and it'd be good to get this in so the PR for the `development` -> `main` branch can get the tests running. Will merge.

Issue #199, opened 2025-12-08
ninja too agressive; reduce default number of threads?


MatthiasHeilManchester:
We need to do something about the way in which ninja can make computers melt. Using the number of cores + 2 (!) for the build process is too much and has resulted in blue screens on various machines. Not likely to endear ourselves to underpowered newbie users.

The required cmake machinery is described here

          https://cmake.org/cmake/help/latest/module/ProcessorCount.html

Should we set the default to  "cores - 2" and/or allow (and explain to) users how to set it to different values?

BTW: Not sure why the example uses logical rather than physical cores; either can be queried. I would have thought it's the physical ones that matter.

  Thoughts?

     Matthias

PR #201, opened 2025-12-10
Update `oomph_build.py`

puneetmatharu:
### Summary 

 - Adds support for `-j`/`--parallel` flag.
 - Default build parallelism to `max(1, num_processors - 2)`.
 - Add `--clean-only` flag to just do a clean and no building.
 - Add info to README on using 
   ```bash
    export CMAKE_BUILD_PARALLEL_LEVEL=$(python3 -c "import os; n=os.cpu_count() or 1; print(max(1, n - 2))")
   ```

Addresses https://github.com/oomph-lib/oomph-lib/issues/199.

(N.B. I'm purposely targeting the `main` branch and ignoring `development`.)

**EDIT: Hmm, I need to come back to this and handle the demo drivers.**

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,9 +58,9 @@
 - [Prerequisites](#prerequisites)
   - [CMake](#cmake)
   - [Ninja](#ninja)
-  - [[macOS only: OpenBLAS]](#macos-only-openblas)
+  - [\[macOS only: OpenBLAS\]](#macos-only-openblas)
   - [Compilers](#compilers)
-  - [Other tools (and how to install all the prerequisites)](#other-tools-and-how-to-install-all-the-prerequisites-in-one-go)
+  - [Other tools (and how to install all the prerequisites in one go)](#other-tools-and-how-to-install-all-the-prerequisites-in-one-go)
 - [Building, installing and uninstalling `oomph-lib`](#building-installing-and-uninstalling-oomph-lib)
   - [Required/optional third-party libraries](#requiredoptional-third-party-libraries)
   - [Step 1: Installing the third-party libraries](#step-1-installing-the-third-party-libraries)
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -201,6 +199,16 @@ sudo apt install ninja-build
 
 will do the trick.
 
+> [!IMPORTANT]
```
MatthiasHeilManchester, line 202: Is this superfluous now that we're doing this by default? We may (!) want to mention what we're doing (and maybe provide a forward reference re how to overrule this and make the build run faster -- if he machine can handle it!) but shouldn't give the impression that here's another big thing to do before you start building.

README.md
```diff
@@ -526,6 +536,16 @@ The `oomph_build.py` script is provided as a convenient one-stop solution for bu
 > [!IMPORTANT]
 > In addition to the standard requirements (CMake and Ninja, as described above), you will need a working Python 3 installation to run `oomph_build.py`. Ensure that `python3` is available in your `PATH`. If you plan to use the script's documentation-building feature (described below), make sure `doxygen` is installed as well.
 
+> [!NOTE]
```
MatthiasHeilManchester, line 539: this is more like it. How about we remove the backward reference (and the bit it points to) and simply highlight the issue (and its resolution) here. The relevant text for that is fine.

external_distributions/CMakeLists.txt
```diff
@@ -131,10 +131,23 @@ include(OomphGetExternalProjectHelper)
 # --------------------------------[ GENERAL ]----------------------------------
 
 # Decide on number of jobs to use for building projects
-ProcessorCount(OOMPH_NUM_JOBS)
+
+# Check CMAKE_BUILD_PARALLEL_LEVEL first and if not defined, defer to checking
```
MatthiasHeilManchester, line 135: good. does what I (chatgpt!) did in my own PR (which I'll now withdraw); what you do is more comprehensive..

puneetmatharu:
Dumping some thoughts here to come back to (hopefully) on the weekend.

> Good. But should we implement the "limit number of processor to use for build" at the cmake level (as done in my PR)? 

I'm hesitant to hack this in everywhere as in most cases it should work fine. Therefore, at least from the CMake side, I think we should (by default) build with the normal amount of jobs that are passed to `make` or `ninja`. Hence why I only set the required jobs argument to CMake from the wrapper script. The problem with the demo drivers is that they can be built at any level which means the hack for the job pool needs to be put in every single file. I need to think more about it.

> While we encourage people to use the python-based build script, many people (incl. myself you'll be pleased to hear) use raw cmake and it would be nice to have the default behaviour be such that it makes crashes less likely. There's a default for the number of cores to be used anyway (invented by ninja; the user isn't particularly aware of this). We just change it (and the user isn't particularly aware of that either), but will appreciate if their machine doesn't die. So "out of the box" you get a robust build process, with a FAQ item, acknowledging the option to make the build process run faster by doing "Blablabla".

This is essentially what the following snippet that I added to the README was meant to address:

> [!IMPORTANT]
> By default, the number of threads Ninja will attempt to N+2, where N is the number of CPU cores.
> On some systems this can result in the system crashing so, unless you are sure that your system
> can handle using this many threads, we recommend that you set the following environment variable
> before running a Ninja build to ensure that the number of threads it uses is equal to N-2:
>
> ```bash
> export CMAKE_BUILD_PARALLEL_LEVEL=$(python3 -c "import os; n=os.cpu_count() or 1; print(max(1, n - 2))")
> ```

Users can set the build parallel level once as an environment variable (possibly in their `.bashrc`) if they want to keep it "forever", and CMake will read it from the environment and only use that many jobs if the variable is set.

> Also: should we wait with the PR until we've got the capability to pass on the information to the self-tests/demo_drivers? That's where Hao and others had real problems, so I'm keen to deal with this in one go before asking people to try it out.

With the environment variable, users can easily set it once when they create their shell if they want to use raw CMake. It's also minimally invasive. If we want to use the other approach, we're going to need to hack the way the CMake builds are done everywhere and then every demo driver project build. It's not really how we should be doing things. I'll try to think on this further and see if specifying the job pool limit is the way to go...

MatthiasHeilManchester:
@puneetmatharu : just noticed that his was still sitting there. Can probably also go. 

BTW: Happy to use the tex suggested above:

> 
> Important
> 
> By default, the number of threads Ninja will attempt to N+2, where N is the number of CPU cores.
> On some systems this can result in the system crashing so, unless you are sure that your system
> can handle using this many threads, we recommend that you set the following environment variable
> before running a Ninja build to ensure that the number of threads it uses is equal to N-2:
> 
> export CMAKE_BUILD_PARALLEL_LEVEL=$(python3 -c "import os; n=os.cpu_count() or 1; print(max(1, n - 2))")

to explain the issue in the README.md file (as suggested in my comments on your latest PR).

MatthiasHeilManchester (commented):
We can probably delete this. Your most recent PR addresses (nearly) all of these issues. Have left a comment to rescue the suggested text for the README.md

README.md
puneetmatharu: It's indeed superfluous if the build is driven by `oomph_build.py`. However, if the user builds the main project (not TPLs) or demo drivers with CMake by hand, using the ancient `cmake` incantations, then they'll need to set this env variable themselves unfortunately. You're absolutely right that I should link to where I mention how to overrule it though

puneetmatharu:
Happy to throw this PR away and update the README in #228!

MatthiasHeilManchester:
Another one hits the bin because @puneetmatharu has done a nicer version of it (not a complaint,btw!) :)

PR #216, opened 2026-03-16
Feature: Solid plastic elements implementing a multiplicative hyperelastic plasticity model

holgergoe:
We've implemented the "Multiplicative Hyperelastic-Based Plastic Equation Incorporating Extended Subloading Surface Model" described by Hashiguchi, K. Multiplicative Hyperelastic-Based Plasticity for Finite Elastoplastic Deformation/Sliding: A Comprehensive Review. Arch Computat Methods Eng 26, 597–637 (2019). https://doi.org/10.1007/s11831-018-9256-5

The file plastic_uniaxial_strain_test.cpp implements a simple uniaxial test case of the element.

Co-authored-by: Thomas Scrase <t.scrase@mercurylab.org>
Co-authored-by: Thomas Scrase <60260897+thomas-scrase@users.noreply.github.com>


src/constitutive/plastic_constitutive_laws.h
```diff
@@ -0,0 +1,554 @@
...
+
+namespace oomph
+{
+
+  /// A base class for isotropic hardening laws to be used / stored by the
+  /// class PlasticConstitutiveLaw.
+  class IsotropicHardeningLaw
+  {
+  public:
+    IsotropicHardeningLaw(double* isotropic_hardening_factor_in_pt)
```
MatthiasHeilManchester: Comments! Explain what the input parameters are (everywhere...)

src/constitutive/plastic_constitutive_laws.h
```diff
@@ -0,0 +1,554 @@
...
+      const DenseMatrix<double>& dSigmavmdM,
+      RankFourTensor<double>& ddSigmavmdMdM) = 0;
+  };
+
+  class VonMisesYieldCriterion : public YieldCriterion
+  {
+  public:
+    VonMisesYieldCriterion() : YieldCriterion() {}
+
+    /*!
```
MatthiasHeilManchester: Please use /// rather than /* */

src/constitutive/plastic_constitutive_laws.h
```diff
@@ -0,0 +1,554 @@
...
+    void surface_function_second_derivative(
+      const double& f,
+      const DenseMatrix<double>& dSigmavmdM,
+      RankFourTensor<double>& ddSigmavmdMdM) override
+    {
+      const unsigned int nrow = dSigmavmdM.nrow();
+      const unsigned int ncol = dSigmavmdM.ncol();
+      ddSigmavmdMdM.resize(nrow, ncol, nrow, ncol);
+
+      if (f < 1e-15)
```
MatthiasHeilManchester, line 313: Really need some explanation of the logic here (and possibly in the comment describing what the function does). Applies to all functions; would also be helpful to refer to equations in the paper that this is all based on so we can cross check, especially if there's no tutorial/theory review to embedded in oomph-lib (which there ideally should be!). We'll have to maintain this in future and without understanding what you've done (possibly not entirely bug free..) we'll be completely lost.

src/constitutive/plastic_constitutive_laws.h
```diff
@@ -0,0 +1,554 @@
...
+      {
+        ddSigmavmdMdM.initialise(0.0);
+        return;
+      }
+
+      // The hessian is
+      // d^2f/(dM_ij dM_kl) = 1 / f *
+      //           [ 3 / 2 (\delta_ik \delta_jl - 1 / 3 \delta_ij \delta_kl)
+      //             - df/dM_kl * df/dM_ij]
+      double invf = 1 / f;
```
MatthiasHeilManchester: 1.0/f

src/constitutive/plastic_constitutive_laws.h
```diff
@@ -0,0 +1,554 @@
...
+                        double* elastic_core_u_in_pt)
+      : normal_yield_ratio_elastic_pt(normal_yield_ratio_elastic_in_pt),
+        u_pt(u_in_pt),
+        elastic_core_u_pt(elastic_core_u_in_pt)
+    {
+    }
+
+
+    NormalYieldRatioLaw(double* normal_yield_ratio_elastic_in_pt,
+                        double* u_in_pt,
```
MatthiasHeilManchester: General question: Does the "_in_" indicate an input parameter? This is confusing and certainly not in line with the general oomph-lib conventions.

src/constitutive/plastic_constitutive_laws.h
```diff
@@ -0,0 +1,554 @@
...
+                        double* elastic_core_u_in_pt,
+                        double* fd_step_in_pt)
+      : normal_yield_ratio_elastic_pt(normal_yield_ratio_elastic_in_pt),
```
MatthiasHeilManchester: Private member data is uppercase as in Normal_yield_ratio_elastic_pt. Applies everywhere!

src/constitutive/plastic_constitutive_laws.h
```diff
@@ -0,0 +1,554 @@
...
+        elastic_core_u_pt(elastic_core_u_in_pt),
+        fd_step_pt(fd_step_in_pt)
+    {
+    }
+
+    double compute_r_plastic(const double& u,
+                             const double& delta_lambda,
+                             const double& R_prev)
+    {
+      double derivative;
```
MatthiasHeilManchester: Initialise everything!

src/constitutive/plastic_constitutive_laws.h
```diff
@@ -0,0 +1,554 @@
...
+
+      // Compute the argument of cos^{-1}
+      // We do not need a max around R_prev - Re, since R_prev >= Re by
+      // definition
+      double inner_arg = std::cos(preFactorArg * (R_prev - Re)) *
+                         std::exp(-u * preFactorArg * delta_lambda);
+
+      // Limit the arg. This has two effects, it prevents the acos to take wierd
+      // numbers and makes the derivative always not infinite.
+      const double TOLERANCE = 1.0e-12;
```
MatthiasHeilManchester: all caps only for template parameters

src/constitutive/plastic_constitutive_laws.h
```diff
@@ -0,0 +1,554 @@
...
+          // dR/du = preFactor * inv_sqrt * dfdu
+          //       = - f * inv_sqrt * delta_lambda
+          dRdu = -inner_arg * inv_sqrt * delta_lambda;
+        }
+      }
+
+      return R;
+    }
+
+    double compute_u_with_elastic_core(
```
MatthiasHeilManchester, line 545: Ouch. the number of arguments is enormous and calling this function will be extremely error prone. is there any way we can pass in the arguments via a struct/class. Not so much of a big deal if this is only called internally (which I can't tell).

src/constitutive/plastic_constitutive_laws.h
```diff
@@ -0,0 +1,554 @@
...
+
+    // Elastic core law
+    ConstitutiveLaw* elastic_core_law_pt;
+
+    // For computing the normal yield ratio
+    NormalYieldRatioLaw* normal_yield_ratio_law_pt;
+
+    double* eta_p_pt;
+
+    /*!
```
MatthiasHeilManchester: We don't need \brief any more. And please change to ///

src/constitutive/plastic_constitutive_laws.h
```diff
@@ -0,0 +1,554 @@
...
+                           const RankFourTensor<double>& dbarbar_N_dbarbar_M,
+                           const double& dbarbar_M_dbar_Mk,
+                           const double& dbarbar_M_dbar_Mc,
+                           const DenseMatrix<double>& dbarbarM_dR,
+                           const RankFourTensor<double>& dhatbar_Nc_dhatbar_Mc,
+                           DenseMatrix<double>& dC_dbar_M,
+                           DenseMatrix<double>& dC_dbar_Mk,
+                           DenseMatrix<double>& dC_dbar_Mc,
+                           double& dCdR,
+                           bool computeDerivative)
```
MatthiasHeilManchester: compute_derivative. No internal capitalisation unless it's a class name. Exeptions are acronyms (like MPI), and cases where it really makes the structure clearer, e.g. if a letter refers to a matrix that one would automatically write in uppercase (so something like dbarbar_N_dbarbar_M may be ok (and is certainly less offensive because it has underscores)

MatthiasHeilManchester (changes requested):
Sorry it's taken so long to have a first look at this. I'm going to be away for 10 days from Monday and wanted to give you something to do while I'm away (in case you get bored!). I've only annotated one file so far and that's already identified a fairly large number of problems that'll also apply to the other files. I know it's a bloody pain to fix these things retrospectively (so I always tell my students to comment as they go along), and some of the requests may seem minor/annoying to you. However, the library will just fall apart if we're no imposing a uniform style. 

We should also discuss (probably when I'm back) how much theory write up we're going to accompany this with. At the moment, this code would be unmaintainable because it's not clear where the equations come from. In-code references to the relevant sections/equations/algorithms in the review paper would already help massively. 

The other big thing is the range of testing: It seems there's only one driver code which does a very simplistic test (uniaxial strain) and thus almost certainly doesn't even remotely touch all the capabilities that these elements are supposed to have. What do people usually use as established validation cases? There must be some! Given how complex the machinery is, it may be difficult to explore/validate it all (properly, by comparing against known solutions), but then the relevant bits of untested (in that sense) code should come with massive disclaimers to warn users. 

  Addressing the marked up comments in the code (there and in the other changed/newly-added) files should be pretty straightforward (if tedious; sorry!) and unless my comments are unclear, just go for it. The rest we'll need to discuss in more detail.

PR #220, opened 2026-03-27
option to delete Validation directories for succesful tests by setting environment …

MatthiasHeilManchester:
…variable OOMPH_DELETE_VALIDATION_DIRECTORY_AFTER_SUCCESSFUL_TEST to "yes". Methodology is straightforward and implemented in (auto-generated run_tests.sh files). Haven't looked at how to set the environment variable from within the github test procedure (which is the main antipated usage, but I trust it can be done. 

If you ( @puneetmatharu ) are happy with this, I will explore and will also mention the option to do this in the FAQ (since the Valdation directories take up a lot of space everywhere).

### DIFF
--- a/cmake/OomphAddTest.cmake
+++ b/cmake/OomphAddTest.cmake
@@ -259,9 +259,23 @@ function(oomph_add_test)
   file(APPEND "${TEST_SCRIPT}" "  exit 1\n")
   file(APPEND "${TEST_SCRIPT}" "fi\n\n")
 
+  # Clean up?
+  file(APPEND "${TEST_SCRIPT}" "# If we're here the test has passed\n")
+  file(APPEND "${TEST_SCRIPT}" "cp ${CMAKE_CURRENT_BINARY_DIR}/Validation/validation.log . \n")
+  file(APPEND "${TEST_SCRIPT}" "dir=\"${CMAKE_CURRENT_BINARY_DIR}/Validation\"\n")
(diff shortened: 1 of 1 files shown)


cmake/OomphAddTest.cmake
```diff
@@ -259,9 +259,23 @@ function(oomph_add_test)
   file(APPEND "${TEST_SCRIPT}" "  exit 1\n")
   file(APPEND "${TEST_SCRIPT}" "fi\n\n")
 
+  # Clean up?
+  file(APPEND "${TEST_SCRIPT}" "# If we're here the test has passed\n")
```
puneetmatharu, line 263: Why not move this block to after the appending to the global `validation.log`? That way you don't have to create a copy of the validation log; you can just print where the global validation log is.

Also I think there might be a bug here introduced by me; if we exit on line 253 because the test failed, we never append to the global log (which gets uploaded by GitHub in the CI); we should make sure this is appended (if the file exists). Do you want to handle that in this PR?

MatthiasHeilManchester: Sure; I'll also have to figure out how to set the relevant environment variable for the github tests and mention the functionality in the FAQ, so theres a bit more work needed anyway. Will do this when w're back from the holiday.

MatthiasHeilManchester: @puneetmatharu : I assume this has all been incorporated (in a non-offensive way, I'm happy to confirm) into your own version of this? If so, I'm happy to delete this PR. (I've just submitted a few minor tweak requests).

puneetmatharu: It is indeed!

MatthiasHeilManchester:
Closing because Puneet has done it in a parallel universe (pull request)

PR #222, opened 2026-04-16
public virtual inheritance from geometric finite elements & tolerance tweak

MatthiasHeilManchester:
- Needed to avoid ambiguities when combining these elements in a PseudoSolid context. 
- demo_drivers/multi_physics/spherical_shell_convection has been very sensitive; bumped tolerance yet again


alhazel (approved):
All looks fine to me. No idea why the spherical convection is so sensitive either. Might be a poor choice of measures of the solution.

PR #229, opened 2026-07-31
dshape hot fix

MatthiasHeilManchester:
Fixed various crimes in the implementation of dshape_and_dtest_eulerian_* Various different types of mistakes appear to have got copied all over the place without thinking. I'm pretty sure I've found all of them and have now replaced the loop based copy operations uniformly with the copy constructors. Bizarrely (and luckily!) none of this affected the self-tests. 

Also updated fake command line argument for inline tetgen from Y to YY to suppress splitting of boundaries.


src/advection_diffusion/advection_diffusion_elements.h
```diff
@@ -727,16 +727,9 @@ namespace oomph
...
-    // shape functions
-    for (unsigned i = 0; i < NNODE_1D; i++)
-    {
-      test[i] = psi[i];
-      for (unsigned j = 0; j < DIM; j++)
-      {
-        dtestdx(i, j) = dpsidx(i, j);
-      }
-    }
+    // Set the test functions equal to the shape functions
```
alhazel: There is a (subtle) distinction here. The equality operator does a shallow copy whereas the loop does a deep copy. This could have unintended consequences higher up, so needs to be clearly flagged somewhere. Discuss?

MatthiasHeilManchester:
@alhazel Good news: you'd already dealt with all the instances where the shallow assignment opeprator was used (yay!), so after merging your changes into my branch I simply deleted your commented-out deep copier and pushed to the repo. Self-tests are running. If they pass, we should be good to go!

MatthiasHeilManchester:
> @alhazel Good news: you'd already dealt with all the instances where the shallow assignment opeprator was used (yay!), so after merging your changes into my branch I simply deleted your commented-out deep copier and pushed to the repo. Self-tests are running. If they pass, we should be good to go!

...all passed!

src/advection_diffusion/advection_diffusion_elements.h
alhazel: Resolved.

alhazel (approved):
All looks good to me. Approved.

PR #232, opened 2026-08-31
Add Nathan's doc directory to top-level CMakeLists.txt file; update Doxyfiles; COLLABORATING_ON_PR.md 

MatthiasHeilManchester:
One minor tweak to the previous PR (which I handled); updated the Doxyfiles too to get rid of warnings. Builds cleanly and self-tests pass.


MatthiasHeilManchester:
I have added a new document COLLABORATING_ON_PR.md that is referenced from CONTRIBUTING.md to explain how submit PRs on PRs. Wouldn't mind @puneetmatharu to give this a quick sanity check.

  Other than that, this should (hopefully!) be really easy to approve -- it's a minor fix to make sure that Nathan's new documentation actually gets built (overlooked recursion in a CMakeLists.txt file). The changes to the Doxyfiles were machine generated by Doxgen itself and get rid of warnings.

COLLABORATING_ON_PR.md
```diff
@@ -0,0 +1,324 @@
...
+   **[Replace `bug_fix` with the name of the actual branch.]**
+- Your own fork of  `oomph-lib`'s GitHub repo is `JosephineMaintainer/oomph-lib`. This represents the GitHub repo at https://github.com/JosephineMaintainer/oomph-lib. 
+
+   **[Replace JosephineMaintainer with your GitHub username.]**
+
+---
+
+## One-time Setup
+
+Check out a clean version of the `oomph-lib` repository
```
alhazel: Technically this is not needed. One can add extra remotes (as many as you like) to an existing repo. Network bandwidth and diskspace is not infinite, so might be better to just add the remote. Would suggest. "Although, we recommend checking out a clean version of the oomph-lib repository this is not necessary. You can add/change remotes on an existing repository. The key point is that you need to be able to access the main oomph-lib repository to fetch the PR branch, but all edits should be done on branches created in your own fork of the library. In other words, your working repository needs (at least) two remotes: the main oomph-lib library and your fork."

alhazel (commented):
Generally happy. Have one tiny (optional) suggestion, but don't feel very strongly about it.

MatthiasHeilManchester:
@alhazel OK; I've run your proposal past my chatty friend and have essentially said what you said (you may disagree!) but we've introduced the word "local clone" rather than using "repository" in too many different meanings, the logic now being that oomph-lib = a repository (on github); your fork  = another repository on github, obtained by forking oomph-lib;  a local clone = whatever you've cloned from github onto your machine.  Does the new version work for you?

